PR #18, opened 2021-04-16
Heart app

diazandr3s:
Created an initial app for heart segmentation. 

I still have some questions regarding this:

- Is it possible for the researcher to ONLY focus on creating/modifying scripts in the app folder? We still need to add an engine in the monailabel/engines/infer
- I see there are many places where the transforms are defined (monailabel/engines/infer, heart_app/lib/infer, heart_app/lib/train). I've got a bit lost modifying the transforms. Don't you think having a "central" file in which the transforms are defined for test and train could be cleaner?
- Is there a way to automatically create the YAML file based on the MONAILabelApp attributes, methods, and/or method arguments? I forgot whether I should modify the YAML file, the MONAILabelApp or both

Sorry if I'm missing somenthing.


SachidanandAlle:
You r right..  Few things in yaml will be automatically added/generated based on your App.. so that you can only do minimum amount work..

And yes u will be only focusing on your app.. you don't have to contribute to monailabel.. as we discussed.. in this one.. you r basically helping to add/bundle heart engines (train/infer) for others to directly use..

Monday.. lets sit together and walk through this code.. and lets see how you can add minimal code for your next app..

SachidanandAlle:
In some cases.. your Transform for train, validation and infer/test can be different.. (example deepgrow).. lets discuss more on Monday

SachidanandAlle:
For inference u don't need network.. convert or export it to torchscript

SachidanandAlle (approved):

SachidanandAlle:
Considering your feedback, I have fixed few things and made it simple.. please checkout main and pull the latest branch

PR #23, opened 2021-04-21
Msd infer engines

diazandr3s:
- Add DeepEdit Engine
- Add Engine for BRATS, Hippocampus, Prostate


monailabel/engines/infer/segmentation_heart.py
```diff
@@ -69,4 +70,4 @@ def post_transforms(self):
             ToNumpyd(keys='pred'),
             Restored(keys='pred', ref_image='image'),
             BoundingBoxd(keys='pred', result='result', bbox='bbox'),
-        ]
+        ]
```
SachidanandAlle, line 73: are u sure this is not by mistake? i mean ] is removed...

PR #25, opened 2021-04-22
Adding Dataset object for data management

aihsani:
## Description

Dataset object used to manage access to the studies/dataset loaded into MONAI Label Server. Dataset object reads folder of images with no `dataset.json` and assumes all files are unlabeled, adding all files it finds in subtree of `--studies` path argument to the `images` category of objects.

### Implemented

Dataset is currently used in the MONAI app class under `save_label` and `next_image`


monailabel/interface/app.py
```diff
@@ -155,12 +155,10 @@ def save_label(self, request):
...
             JSON containing next image and label info
         """
-        # TODO:: Save label, trigger training (if condition is met)
-        labels_dir = os.path.join(self.studies, "newLabelsTr")
-        os.makedirs(labels_dir, exist_ok=True)
-
-        label_file = os.path.join(labels_dir, os.path.basename(request["image"]))
-        shutil.move(request["label"], label_file)
+        
+        label = io.BytesIO(open(request['label'], 'rb').read())
```
SachidanandAlle, line 159: for now it's ok.. 
endpoint can send the FP for the api to read it from there.. we can avoid multiple writes for the label file

SachidanandAlle (approved):
Looks good to me.. we can squash and merge

SachidanandAlle:
tried with the changes.. next sample and save label works fine..
one suggestion.. for labelname lets have some part of it to represent image name while saving.. 
may be label_<image_file_name>.<extension>

monailabel/interface/app.py
aihsani: Sounds good, I was going to suggest we keep the file in memory if possible so I save it only once here.

aihsani:
Do you want me to update to list under images anything that starts with `image_` and add `label_` to saved labels?

My assumption was that perhaps if the dataset is entirely not annotated (missing a dataset.json) then anything found in the subtree is images. The reason I didn't put limitations on files names or paths is because from experience I know it's painful to update hundreds of files to match a certain format. We can discuss what's the best way, but your suggestions is good so we can do that short term.

SachidanandAlle:
have fixed name.. also added a small util to get the datalist (for training) over existing labels
https://github.com/SachidanandAlle/MONAI-label/commit/bbc665d6e6884e9f6a5981210e51b0797f824d8c

Issue #42, opened 2021-05-04
Create and use user-app-specific virtual environment for training


aihsani:
Problem: When creating a training session for the user app we need to separate the libraries required by the user app from MONAILabel

PR #43, opened 2021-05-04
Setting up and using  user-app virtual environments for training

aihsani:
Address issue #42

Changes:
- moved `monai-label` specific dependencies outside the user app code branch
- added shell script to allow running a training session/process

Impact:
Changes impact end-users as they will need to now install MONAILabel and run it as a command line client in order to be able to have access to the shell scripts required to run the training branch of the user-app. Specifically, the user will need to:
```bash
git clone https://github.com/Project-MONAI/MONAILabel.git
cd MONAILabel
pip install .
```
to install `monai-label`, and run the MONAILabel server as
```
monai-label --app <path_to_app> --studies <path_to_data>
```


monailabel/interfaces/tasks.py
```diff
@@ -24,17 +24,12 @@
 
 
 def task_func(task, method):
-    import monailabel.utils.others.app_utils as app_utils
     cmd = [
-        "python3",
-        os.path.realpath(app_utils.__file__),
-        "-a",
+        "run_monailabel_user_app.sh",
```
SachidanandAlle, line 28: two options
1) set PATH to scripts folder (or pip install will add them bin), but howabout dev env (without pip)
2) scripts folders part of pip package and you can reference run_monailabel_user_app.sh the based on current python file path

SachidanandAlle:
Except one comment.. to enable working in dev env (without pip install).. 
Rest of the function/logic looks good to me.. you can merge to main brain

monailabel/interfaces/tasks.py
SachidanandAlle: normally something like this.. can take care of if running in dev env vs pip mode

```
cmd_script = os.path.realpath(os.path.join(tasks.__file__, '..', 'scripts', 'run_monailabel_user_app.sh'))
cmd_script = cmd_script if os.path.exists(cmd_script) else 'run_monailabel_user_app.sh'
```

monailabel/scripts/run_monailabel_user_app.sh
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
```
SachidanandAlle, line 1: > since we use direct file to run the command (instead of sh <filename>), when you git commit, please make sure you have +x permissions for this file.

monailabel/interfaces/tasks.py
aihsani: There's a second issue here which we can discuss, as the `run_monailabel_user_app.sh` contains a `monai-label-utils` command [here](https://github.com/Project-MONAI/MONAILabel/blob/aihsani/training_venv/monailabel/scripts/run_monailabel_user_app.sh#L42).

We could perhaps add a flag that executes a "developmemt" branch of this script

SachidanandAlle: The script file (internal modules etc..) can always refer full path instead of entry point.. 

Entry points through pip are mainly for external or package users

SachidanandAlle (approved):

PR #54, opened 2021-05-05
Create workflow for DeepEdit

diazandr3s:
- Change the way transforms are defined in **sample-apps/segmentation_heart/lib/infer.** Add Compose to transforms defintion
- Implement the inverse transform for inference. See **monailabel/utils/infer/task.py**
- DeepEdit is working for heart images. Only inference is working for Spleen image. Working on the DeepGrow functionality for Spleen

### DIFF
--- a/monailabel/utils/infer/deepgrow_3d.py
+++ b/monailabel/utils/infer/deepgrow_3d.py
@@ -1,8 +1,9 @@
+import numpy as np
 from monai.apps.deepgrow.transforms import (
     AddGuidanceFromPointsd,
     SpatialCropGuidanced,
-    ResizeGuidanced,
     AddGuidanceSignald,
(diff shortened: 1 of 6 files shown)


SachidanandAlle:
Can we create a separate app for deepedit.. say for now one organ..  and then we can easily flip abd decide which one to include in the release

diazandr3s:
> Can we create a separate app for deepedit.. say for now one organ.. and then we can easily flip to decide which one to include in the release
> 
> If we are guaranteed to release deepedit only then we can override it here.. and ignore my above comment

It is a good idea! The only issue I see of creating a separated app for deepedit is that we still need to modified the **monailabel/utils/infer/task.py**

SachidanandAlle:
We don't need this anymore.. Lets close the PR

PR #96, opened 2021-05-27
Batch Inference

aihsani:
Enabling batch inference of unlabeled images for (automated) segmentation models loaded in the user-app.

The aim is to ultimately allow the user to derive uncertainty scores from the inference results of the batch inference.


aihsani:
Sachi, I think one outstanding item here that we might need to gather requirements for is how we're going to enable the user app to  create scores from the generated labels in the batch inference

Issue #97, opened 2021-05-27
Training/Batching all 2D slices for Deepgrow 2D while training


finalelement:
Currently, during training only a single slice is chosen randomly from a given 3D volume. This process while suffices for training, it needs to be improved. The improvement can be done by selecting multiple slices. This will also allow for a higher batch size during training.

Possible Solutions:
1.) We extract the batch size number of slices randomly from a 3D Volume and use that for training.

2.) We flatten an entire list of json 2D slices and cover that for training

PR #136, opened 2021-06-16
Flatten 2D based on config + add train/val batch size config

SachidanandAlle:
- Flattend 2D train/val datalist for deepgrow 2D apps
- Add train/val batch size config for all apps
- In Slicer, use current/total epoch from train stats instead of locating from train logs


SachidanandAlle:
https://github.com/Project-MONAI/MONAILabel/issues/97
Flatten based on number of samples to pickup (without image/label read) for each 3D volume.. make it a config item

Seperate epoch/batch config for 2D model in info.yaml for deepgrow apps

PR #138, opened 2021-06-20
Add pretrained flag

diazandr3s:
Add the flag "pretrained" so users can do cold-start training without many changes. 

- How to manage error when a user tries to perform inference without cold-start training and without having a pretrained model activate (pretrained flag is False)
- Should we do the same for DeepGrow? @finalelement 


SachidanandAlle:
CI Check failed..

diazandr3s:
Thanks! It should be ready now

> CI Check failed..

PR #148, opened 2021-06-28
Release workflow fixes

aihsani:
- PyPI test publication and artifact publication should now function correctly based on tags


SachidanandAlle (approved):
LGTM

PR #155, opened 2021-06-29
Inference integration test for autosegmentation of heart ventricle

aihsani:
- Added integration test for segmentation inference
- Made monai label server port configurable via `MONAILABEL_SERVER_PORT` environment variable
  - This is useful for local testing
- Fixed non-terminating server when integration tests fail

### DIFF
--- a/tests/integration/test_infer.py
+++ b/tests/integration/test_infer.py
@@ -1,0 +1,19 @@
+import unittest
+
+import requests
+
+from . import SERVER_URI
+
+
+class EndPointInfer(unittest.TestCase):
+
+    def test_segmentation(self):
+        model = "segmentation_left_atrium"
+        image = "la_004.nii.gz"
+
+        response = requests.post(f"{SERVER_URI}/infer/{model}?image={image}")
+        assert response.status_code == 200
+
+
+if __name__ == "__main__":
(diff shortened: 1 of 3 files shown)


tests/integration/test_infer.py
```diff
@@ -0,0 +1,18 @@
+import unittest
+
+import requests
+
+from . import SERVER_URI
+
+
+class EndPointInfer(unittest.TestCase):
+    def test_segmentation(self):
+        model = "segmentation_left_atrium"
```
SachidanandAlle, line 10: lets have gpu check.. and skip if gpu is not enabled kind..
```
        if not torch.cuda.is_available():
            return
```

aihsani: why? this can happen on CPU even if it takes longer. it's just inference

SachidanandAlle: K.. some models don't work due to specific network/op not available on cpu.. we can merge if this run for both

SachidanandAlle (approved):

tests/integration/test_infer.py
aihsani: I see. OK. Changing

PR #157, opened 2021-06-30
Example for multilabel app (liver + tumor)

SachidanandAlle:
- liver and tumor app
- fix active learning strategy selection


diazandr3s (approved):
Awesome!

PR #170, opened 2021-07-02
Updated Spleen Segmentation App with MMAR Support (TBD)

aihsani:
- Added model path finder to superclass as that is missing from MONAI in 0.6.0rc1 (filed https://github.com/Project-MONAI/MONAI/issues/2512 to have this created upstream in MONAI)
- Modified the segmentation apps to use `load_from_mmar` to load the pretrained segmentation model


aihsani:
@SachidanandAlle please provide feedbak if this is the direction we want to go or not

SachidanandAlle:
Lets migrate only this app (segmentation_spleen) to use clara mmar (for reference implementation)

sample-apps/segmentation_spleen/main.py
```diff
@@ -14,18 +13,13 @@
...
-        self.network = UNet(
-            dimensions=3,
-            in_channels=1,
-            out_channels=2,
-            channels=(16, 32, 64, 128, 256),
-            strides=(2, 2, 2, 2),
-            num_res_units=2,
-            norm=Norm.BATCH,
-        )
+        self.network = load_from_mmar(self.mmar_model_name, mmar_dir=self.model_dir, map_location="cuda")
```
wyli, line 18: This self.network is already created by using the pretrained weights after line 18, perhaps “self.pretrained_model” is not needed? This mmar caching path should be internal impl. details maintained by monai, please avoid accessing it at this application level if possible

aihsani: In the apps we use the `pt` file saved in the disk to be able to perform training and batch inference in separate processes, and as a way to use the results of training before training is complete. So we need a checkpoint file; `self.network` is not sufficient.

So I have two questions:
1. Are you saying we shouldn't modify the model in the downloaded MMAR, but rather save in a separate path and work from there, never having to update the model in the MMAR?
2. Since an MMAR is being downloaded, should this sample app load the transformations form the MMAR? That is to say, is there a facility in MONAI that allows the loading of transformations from `config_infer.json`?

wyli: For 1, yes, mmar is a read-only file, so that every time we load it, we get the same pretrained net instance. training/validation model updates shouldn't update the checkpoint file in the mmar folder. I'm writing a tutorial notebook for loading mmar. Sorry for the confusion.

For 2, that is not implemented yet. The current mmar loading is mainly following the idea of torch.hub module https://pytorch.org/hub/

SachidanandAlle:
lets hold this usecase for monailabel.. 
for now, we can make sure 0.6 monai version works well for monailabel.
we check all basic features working in monailabel

lets do direct wget for ngc models for now as we want to update them over active learning solution.
we can revisit the exact api use-case later..

SachidanandAlle:
These changes are added/moved in another PR.. 
https://github.com/Project-MONAI/MONAILabel/pull/174

load_mmar API is used to train/infer for following samples
- segmentation_spleen
- segmentation_liver_and_tumor
- deepgrow apps

PR #181, opened 2021-07-08
make labels optional in info.yaml

SachidanandAlle:

If **not provided in info.yaml** it will aggregate all the labels from infer tasks
Otherwise it will continue to use what is provided in info.yaml

That way by default user doesn't have to worry about copying labels into info.yaml (unless he/she wants to provide explicitly in info.yaml)


diazandr3s (approved):
Another step to simplifying the App development :)

PR #217, opened 2021-07-15
Add label outline/contour button

diazandr3s:

- Add button in UI to toggle contour/label outline
- Create connections to toggle contour


plugins/slicer/MONAILabel/MONAILabel.py
```diff
@@ -1002,6 +1008,9 @@ def onSaveLabel(self):
 
         logging.info("Time consumed by save label: {0:3.1f}".format(time.time() - start))
 
+    def onClickContour(self):
+        self.editUtil.toggleLabelOutline()
```
diazandr3s, line 1012: @lassoan I'd like to toggle the contour/label outline on the segmentation nodes in the scene. Somehow it is not toggling the contours. Am I missing something? I'm trying to follow the code available here: https://github.com/Slicer/Slicer/blob/master/Modules/Scripted/EditorLib/EditUtil.py#L268

lassoan: EditorLib is for the old Editor module, for labelmaps, not for segmentations. 

You can change segmentation visualization options in the display node (vtkMRMLSegmentationDisplayNode) of the segmentation node that you can get by calling segmentationNode.GetDisplayNode().

diazandr3s: Excellent! Thanks @lassoan !

plugins/slicer/MONAILabel/MONAILabel.py
```diff
@@ -169,6 +170,7 @@ def setup(self):
         self.ui.serverComboBox.lineEdit().setPlaceholderText("enter server address or leave empty to use default")
         self.ui.fetchServerInfoButton.setIcon(self.icon("refresh-icon.png"))
         self.ui.segmentationButton.setIcon(self.icon("segment.png"))
+        self.ui.contourButton.setIcon(self.icon("icon_contour.svg"))
```
SachidanandAlle, line 173: Can we rename this file to contour.svg to make all same similar.. already they are in Icons folder 

Also add this in CMAKE file.. to be part of release plugin

plugins/slicer/MONAILabel/Resources/UI/MONAILabel.ui
```diff
@@ -346,6 +333,32 @@
         </property>
        </widget>
       </item>
+      <item row="1" column="1">
+       <widget class="QComboBox" name="segmentationModelSelector">
```
SachidanandAlle, line 337: Do you need this as generic or applicable only for auto segmentation mask.. how about mask from deepedit..
If it is generic and independent.. you can add this under tools section instead of auto segmentation section

SachidanandAlle (changes requested):

plugins/slicer/MONAILabel/Resources/UI/MONAILabel.ui
diazandr3s: Agree!

SachidanandAlle (approved):

PR #270, opened 2021-08-12
Slicer Plugin - fix readme and ui label text for accuracy

SachidanandAlle:

Based on feedback
- [x] Add a note to handle with any previous version for slicer
- [x] Rename name from Model Accuracy => Train Accuracy and move it under Training status

### DIFF
--- a/plugins/slicer/README.md
+++ b/plugins/slicer/README.md
@@ -8,6 +8,7 @@ Pick one of the following options to install MONAILabel Plugin for 3D Slicer
 
 ### Install 3D Slicer Preview Version with in-built Plugin
 
+- **Uninstall** any existing 3D-Slicer version
 - Download and Install [3D Slicer](https://download.slicer.org/) **Preview version**
 - Go to **View** -> **Extension Manager** -> **Active Learning** -> **MONAI Label**
 - Install MONAI Label plugin
--- a/plugins/slicer/README.md
+++ b/plugins/slicer/README.md
@@ -8,7 +8,6 @@ Pick one of the following options to install MONAILabel Plugin for 3D Slicer
 
 ### Install 3D Slicer Preview Version with in-built Plugin
 
-- **Uninstall** any existing 3D-Slicer version
 - Download and Install [3D Slicer](https://download.slicer.org/) **Preview version**
 - Go to **View** -> **Extension Manager** -> **Active Learning** -> **MONAI Label**
 - Install MONAI Label plugin
--- a/README.md
+++ b/README.md
@@ -50,7 +50,8 @@ For **_prerequisites_**, other installation methods (using the default GitHub br
 
 ### 3D Slicer
 
-Download Preview Release from https://download.slicer.org/ and install MONAI Label plugin from Slicer Extension Manager.
+Download **Preview Release** from https://download.slicer.org/ and install MONAI Label plugin from Slicer Extension Manager.
+> **Uninstall** any existing 3D-Slicer version before using the latest preview release. 
 
 Refer [3D Slicer plugin](plugins/slicer) for other options to install and run MONAI Label plugin in 3D Slicer.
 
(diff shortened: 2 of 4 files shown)


plugins/slicer/README.md
```diff
@@ -8,6 +8,7 @@ Pick one of the following options to install MONAILabel Plugin for 3D Slicer
 
 ### Install 3D Slicer Preview Version with in-built Plugin
 
+- **Uninstall** any existing 3D-Slicer version
```
lassoan, line 11: It is completely valid to keep multiple versions of Slicer around (e.g., latest Slicer Stable Release and a few recent Slicer Preview Releases). Also, Slicer does not have to be installed, it is fully portable and can be run from a USB stick 
- should the user not plug in a USB stick that contains Slicer? What is the reason for recommending users to uninstall other Slicer versions?

SachidanandAlle: this is based on the recent complaint from users.. the latest plugin for monailabel is available only in latest preview version.. i guess unlike stable release.. there is no central repo for preview versions.. if people can still pull monailabel plugin with older preview version of 3d slicer.. am i correct?

SachidanandAlle: image

One such situation.. and user had to install previous version and download+install again

lassoan: Stable and preview releases are stored in the same repository and they can be all downloaded from https://download.slicer.org/.

We can tell users that MONAILabel requries Slicer Preview Release revision _NNNN_ or later.

Slicer releases are all installed in different folders by default and the only thing that is shared between them is the Slicer.ini file (version-specific settings are stored in a separate settings file in the application folder), so having multiple Slicer versions installed or present in some folders on the computer should not cause any issues.

lassoan: > One such situation.. and user had to uninstall previous version and download+install again to fix this kind of situation

The error the user reported is due to that we switched to our new Slicer Extensions Manager Server yesterday and a bug came up that was not found during testing (see details [here](https://discourse.slicer.org/t/unable-to-install-extension-incompatible-with-slicer-r30110/19158/2)). The issue is fixed now.

In general, there is no need to uninstall any Slicer version if something is wrong - the can install older or newer releases of the application and they should live without problems side by side. Of course the user can decide to uninstall any Slicer versions that he no longer needs.

SachidanandAlle: Some users are really getting confused.. they are not pro.. they have no idea about versions.. and how to work with multiple versions of slicer.. so thought of adding one simple step for those kind.. 


How about if user wants to get the latest changes for monailabel plugin?  Is it not required to install the latest preview release?  Can I still use a week ago preview version.. and only update the monailabel plugin to pull today's changes?

may be i can update the note accordingly

lassoan: > How about if user wants to get the latest changes for monailabel plugin? Is it not required to install the latest preview release? Can I still use a week ago preview version.. and only update the monailabel plugin to pull today's changes?

Yes, users need to install the latest preview release to get latest monailabel plugin. But they don't have to uninstall previous versions. That is an extra, unnecessary step.

On macOS all Slicer versions are just created by the name "Slicer", and when users install it then they are offered to overwrite the previous version or add it (then the new version will get the name Slicer_1 or something like that). Either option is fine. Again, they don't have to uninstall anything.

SachidanandAlle: Fixed the note.. it shall be only a hint for those normal users.. who doesn't know anything about managing multiple version

lassoan: The current text is much better because it does not require the users to uninstall previous Slicer versions, just recommends it. 

Instead of saying "To avoid conflicts with older version" it would be more accurate to say "To avoid accidentally using an older Slicer version", because there are no conflicts between different Slicer versions. But since all these will go away in a few weeks, with the new Slicer Stable Release, the currently proposed text is OK for this short time.

SachidanandAlle: Updated the comment.. thanks for the inputs :)

diazandr3s (approved):

PR #308, opened 2021-08-27
Adding coldstart option for scribbles-based interactive segmentation

### LINKED ISSUE #265: Scribbles - Don't depend on logits (previous segmentation model)
It shall be great if scribbles work independently of previous model output.
Currently it depends on logits collected through previous segmentation action.

It will be ideal if we can make it work based on 
- image
- label (current label from client)
- scribbles (forground/background corrections)

masadcv:
This is now ready to be reviewed/merged. 

This PR adds a coldstart method for scribbles, this method ignores any logits/model output. Instead it creates a simple histogram-based likelihood model, which is then used to do scribble-based segmentation. In practice this works very well for organs that have clear delineation. It can be improved in a future update when we add  bounding box for initial selection of ROI.

Fixes #265


masadcv:
This is ready, the coldstart code is going in spleen segmentation scribbles app for now. 

I will submit a separate pull request to move common code into ./monailabel/scribbles and another one to move this and other scirbbles methods into a generic segmentation app.

SachidanandAlle (approved):
This is fine for now.. lets think how to pick things and move them to generic section.. which can be added into all generic apps.. in another PR

PR #321, opened 2021-09-02
Move scribbles to generic segmentation app

masadcv:
Fixes #312 

Moving scribbles methods to generic segmentation app, a followup pull request may be required to move spleen_segmentation_scribbles app to sample apps repository.

PR #325, opened 2021-09-02
Add ROI select method to define a boundingbox for coldstart  scribbles methods

### LINKED ISSUE #319: Scribbles - Add ROI select feature to prune background for coldstart
This is a nice-to-have feature request.
At the moment scribbles coldstart requires a lot of user interaction especially to indicate all background regions.

Using an ROI, will enable putting everything outside as background and hence improve the histogram-based likelihood model used within this approach.

This possibly requires the use of vtkMRMLAnnotationROINode for selecting ROI and then passing this as an extra argument from client to relevant transforms at server.

masadcv:
Fixes #319 

This pull request adds the following functionality:
- [x] Implement recording boundingbox ROI in 3D Slicer extension using `vtkMRMLAnnotationROINode`
- [x] Implement conversion of bbox ROI parameters from RAS to IJK space, in order to use these in image volume space
- [x] Package bbox ROI parameters (when avaiable) and send to server as additional parameters
- [x] On server side, implement a transform that looks for ROI parameters and adds background scribbles to all areas outside the bbox
- [x] Use the method to improve interaction for scribbles-based coldstart option
- [x] Test ROI-based coldstart with different samples


masadcv:
This is now ready for review/merge. It may close the original request in #319 , however there are some nice suggestions in that issue that we may want to revisit sometime soon.

masadcv:
Forgot to mention, this is based on code from https://github.com/Project-MONAI/MONAILabel/pull/321 pull requests, so it may be helpful to review this once that has been merged and this branch has been synced with main.

monailabel/scribbles/transforms.py
```diff
@@ -77,6 +77,54 @@ def _copy_affine(self, d, src, dst):
...
+
+    def __call__(self, data):
+        d = dict(data)
+
+        # read relevant terms from data
+        scribbles = self._fetch_data(d, self.scribbles)
+
+        # get any existing roi information and apply it to scribbles, skip otherwise
+        selected_roi = d.get(self.roi_key, None)
+        if selected_roi is not None and selected_roi != []:
```
SachidanandAlle, line 107: `if not selected_roi` should cover both None and Empty array checks

monailabel/scribbles/utils.py
```diff
@@ -126,10 +126,10 @@ def get_values(_data, _seed, _label):
         return _values
 
     values = get_values(image, scrib, scribbles_bg_label)
-    bg_hist, bg_bin_edges = np.histogram(values, bins=512, range=(0, 1), density=True)
+    bg_hist, bg_bin_edges = np.histogram(values, bins=32, range=(0, 1), density=True)
```
SachidanandAlle, line 129: bins i would prefer to keep it in function args with default=32 in this case.. so that while using depending on the use-case, caller can decide

plugins/slicer/MONAILabel/MONAILabel.py
```diff
@@ -344,12 +344,17 @@ def onUpdateScribbles(self):
             self.updateServerSettings()
             self.reportProgress(60)
 
+            roiNode = slicer.mrmlScene.GetFirstNodeByClass("vtkMRMLAnnotationROINode")
```
SachidanandAlle, line 347: Assuming first node of type ROI?  Why don't u have a drop-down in the scribbles.. that can list all the ROI nodes.. and by default first node is selected.. in case user has added multiple ROI nodes.. he has the choice to fix/correct the selection while running the scribbles..

sample-apps/generic_segmentation/main.py
```diff
@@ -12,7 +12,7 @@
 import logging
 import os
 
-from lib import MyInfer, MyStrategy, MyTrain
+from lib import GenericISegGraphCut, GenericISegGraphcutColdstart, GenericISegSimpleCRF, MyInfer, MyStrategy, MyTrain
```
SachidanandAlle, line 15: may be this needs to be merge to main branch first.. ?  i think these changes are part of previous PR which is already merged

masadcv: Yes, you are right. this builds on top of previous PR. Just merged this with main and resolved pending conflicts.

monailabel/scribbles/transforms.py
masadcv: Thanks, I will update the code.

monailabel/scribbles/utils.py
masadcv: Sure, I will update to have this in the code.

plugins/slicer/MONAILabel/MONAILabel.py
masadcv: Yes, thats right! This currently only looks for the first ROI node. The reason for this choice was primarily to keep the code simple while improving the interaction with the usage of ROI and secondly to avoid further cluttering the UI. At the moment the purpose of ROI is only to avoid having the user giving a lot of background scribbles for building a good likelihood model.

It certainly is doable to add additional ROI and have user select the right one. I may need to look into this further and come up with the correct way of highlighting the selected ROI as well during selection.  The rest of the backend logic will remain same.

SachidanandAlle: that's fine... you can make it as selectable list in UI later

SachidanandAlle (approved):

plugins/slicer/MONAILabel/MONAILabel.py
masadcv: Many thanks! Will add this to my backlog and revisit soon.

PR #342, opened 2021-09-08
Scribbles - Copy WriteLogits functionality to generic segmentation app for non-coldstart scribbles methods

masadcv:
With the recent moving of scribbles functionality into generic segmentation app, I seem to have missed the WriteLogits functionality needed for non-coldstart scribbles options. 

This PR attempts to fix this.

- [x] Adding WriteLogits transform to Inferer
- [x] Adding additional code to main.py for handling datastore queries for saving/loading logits
- [x] Tested all three scribbles functionalities within the generic segmentation apps, this includes  1) coldstart, 2) graphcut + logits and 3) crf + logits 

### DIFF
--- a/sample-apps/segmentation/main.py
+++ b/sample-apps/segmentation/main.py
@@ -10,15 +10,22 @@
 # limitations under the License.
 
 import logging
-from monailabel.interfaces.tasks.infer import InferType
 import os
 from distutils.util import strtobool
 
-from lib import GenericISegGraphCut, GenericISegGraphcutColdstart, GenericISegSimpleCRF, MyInferWithWriteLogits, MyStrategy, MyTrain
+from lib import (
+    GenericISegGraphCut,
+    GenericISegGraphcutColdstart,
+    GenericISegSimpleCRF,
+    MyInferWithWriteLogits,
+    MyStrategy,
+    MyTrain,
+)
 from monai.networks.layers import Norm
 from monai.networks.nets import UNet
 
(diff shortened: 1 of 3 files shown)


sample-apps/segmentation/main.py
```diff
@@ -80,3 +88,27 @@ def init_strategies(self):
             "random": Random(),
             "first": MyStrategy(),
         }
+
```
SachidanandAlle, line 91: Lets discuss... probably we want to avoid this for public release.. and keep all generic scribbles for all apps that doesn't have app specific dependency (in this case infer output from other infer tasks)

masadcv: Sure, lets discuss in our meeting. This infer function works just fine for non scribbles tasks. The dependency is broken when we try to run scribbles method that need logits from CNN. Maybe there is a more elegant way of achieving this....

masadcv: Another thought, maybe we can only keep coldstart for this generic app - which wont require logits. Then perhaps we may want to keep atleast one app around that covers/showcases all scribbles functionality (e.g. **segmentation_spleen_scribbles** that is due to be removed by end of this week)

masadcv:
As discussed in the weekly meeting, we are now going to remove all model-based scribbles methods and only keep model-free (coldstart) option for release 0.2. Therefore I am going to remove this and other methods for now and revisit after the release.

PR #346, opened 2021-09-09
Scribbles - Remove model-based scribbles methods from generic segmentation app

masadcv:
This PR removes model-based scribbles method and only keeps model-free (coldstart) scribbles method as discussed in the weekly status meeting and here: https://github.com/Project-MONAI/MONAILabel/pull/342#discussion_r705009954

The generic app now only has mode-free graphcut method which can be used as shown in this short demo: https://www.youtube.com/watch?v=yhC27V3PvIQ

- [x] Renamed coldstart --> HIstogramBasedGraphCut throughout the apps
- [x] Removed model-based scribbles tasks for ISeg+Graphcut, ISeg+SimpleCRF, ISeg+MONAICRF, ISeg+InteractiveGraphcut
- [x] Moved scribbles `HistogramBasedGraphCut` code under ./monailabel/scribbles/infer.py
- [x] Added `HistogramBasedGraphCut` scribbles method to all current sample apps including 1) deepedit, 2) deepgrow, 3) deepgrow_left_atrium, 4) generic_segmentation, 5) segmentation, 6) segmentation_left_atrium, 7) segmentation_spleen and 8) segmentation_spleen_scribbles
- [x] Tested the only remaining scribbles method: HIstogramBasedGraphCut with an example
- [x] (sanity check) Tested scribbles method with one simple example in all apps 1) deepedit, 2) deepgrow, 3) deepgrow_left_atrium, 4) generic_segmentation, 5) segmentation, 6) segmentation_left_atrium, 7) segmentation_spleen and 8) segmentation_spleen_scribbles


sample-apps/segmentation/lib/scribbles.py
```diff
@@ -14,113 +14,16 @@
...
-    ApplySimpleCRFOptimisationd,
     MakeISegUnaryd,
     MakeLikelihoodFromScribblesHistogramd,
-    SoftenProbSoftmax,
 )
 from monailabel.utils.others.post import BoundingBoxd, Restored
 
 
-class ScribblesInferTask(InferTask):
+class GenericISegGraphcutModelFree(InferTask):
```
SachidanandAlle, line 24: lets move this class to https://github.com/Project-MONAI/MONAILabel/tree/main/monailabel/scribbles/
and all other sample apps can just use it.. i believe this is generic enough for all other apps to use?

sample-apps/segmentation/main.py
```diff
@@ -59,9 +59,7 @@ def __init__(self, app_dir, studies, conf):
     def init_infers(self):
         infers = {
             "segmentation": MyInfer([self.pretrained_model, self.final_model], self.network),
-            "Coldstart->ISeg+GraphCut": GenericISegGraphcutColdstart(),
-            "ISeg+GraphCut": GenericISegGraphCut(),
-            "ISeg+SimpleCRF": GenericISegSimpleCRF(),
+            "ModelFreeGraphCut": GenericISegGraphcutModelFree(),
```
SachidanandAlle, line 62: lets rename this to **FreeGraphCut**

tvercaut: What about HistogramBasedGraphCut? There is always a model, in this case a histogram if I'm not mistaken...

sample-apps/segmentation/lib/scribbles.py
masadcv: @SachidanandAlle this has been done now. I dont have resources to test deepedit/deepgrow apps as it seems to be running some TTA. If you have those setup, can you just load a sample and click update in scribbles (no need to add any scribbles for this test) to do a sanity check?

sample-apps/segmentation/main.py
masadcv: I like HIstogramBasedGraphCut, will update the apps. I think here by model-free we meant it does not need output from an external model (as it builds its own during execution).

SachidanandAlle (approved):

PR #371, opened 2021-09-15
Adding unittests for scribbles transforms in ./monailabel/scribbles/transforms.py

### LINKED ISSUE #324: Scribbles - Add unittests for scribbles transforms under ./monailabel/scribbles/transforms.py
As discussed in status meeting, this is to discuss and track progress of unittesting for scribbles transforms. 

@SachidanandAlle, @aihsani when I get to this task,  I may need to discuss for some guidance on how to approach this. At the moment, this is only focused  on scribbles transforms as these are core functions used for scribbles.

masadcv:
Fixes #324 
Adding unittests for the following scribbles transfroms from ./monailabel/scribbles/transforms.py
- `AddBackgroundScribblesFromROId`
- `ApplyGraphCutOptimisationd`
- `ApplyISegGraphCutPostProcd`
- `ApplySimpleCRFOptimisationd`
- ` MakeISegUnaryd`
- `MakeLikelihoodFromScribblesHistogramd`

Since these transforms use all methods currently implemented in ./monailabel/scribbles/utils.py therefore it indirectly covers those functions as well.

Inside the tests, I implemented a simple synthetic data generation function (based on monai's synthetic data functions) - that "simulates" the input data. This is used throughout the tests to verify the mathematical correctness of implemented methods. 

These tests cover different aspects of the above transforms, these include checking:
- 2D/3D input/output data variation
- Input arguments
- Output shapes
- Output comparison with expected output (checking mathematical correctness of algorithm applied)

Checklist:
- [x] All tests passing locally

### DIFF
--- a/tests/unit/scribbles/test_transforms.py
+++ b/tests/unit/scribbles/test_transforms.py
@@ -242,7 +242,7 @@ class TestScribblesTransforms(unittest.TestCase):
         ] = True
         input_data[~mask] = input_param["scribbles_bg_label"]
 
-        np.testing.assert_allclose(input_data, result["scribbles"], rtol=1e-4)
+        np.testing.assert_equal(input_data, result["scribbles"])
         self.assertTupleEqual(expected_shape, result["scribbles"].shape)
         self.assertTupleEqual(test_input["scribbles"].shape, result["scribbles"].shape)
 
@@ -259,7 +259,7 @@ class TestScribblesTransforms(unittest.TestCase):
         ] = True
         input_data[~mask] = input_param["scribbles_bg_label"]
 
-        np.testing.assert_allclose(input_data, result["scribbles"], rtol=1e-4)
+        np.testing.assert_equal(input_data, result["scribbles"])
         self.assertTupleEqual(expected_shape, result["scribbles"].shape)
         self.assertTupleEqual(test_input["scribbles"].shape, result["scribbles"].shape)
 
--- a/tests/unit/scribbles/test_transforms.py
+++ b/tests/unit/scribbles/test_transforms.py
@@ -1,0 +1,292 @@
+# Copyright 2020 - 2021 MONAI Consortium
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import unittest
+
+import numpy as np
+from parameterized import parameterized
+
+from monailabel.scribbles.transforms import (
+    AddBackgroundScribblesFromROId,
+    ApplyGraphCutOptimisationd,
+    ApplyISegGraphCutPostProcd,
+    ApplySimpleCRFOptimisationd,
+    MakeISegUnaryd,
+    MakeLikelihoodFromScribblesHistogramd,
+)
+
+
+def generate_synthetic_binary_segmentation(height, width, num_circles=10, r_min=10, r_max=100, random_state=None):
+    # function based on:
+    # https://github.com/Project-MONAI/MONAI/blob/dev/monai/data/synthetic.py
+
+    if r_min > r_max:
+        raise ValueError("r_min cannot be greater than r_max")
+
+    min_size = min(height, width)
+    if 2 * r_max > min_size:
+        raise ValueError("2 * r_max cannot be greater than min side")
+
+    rs: np.random.RandomState = np.random.random.__self__ if random_state is None else random_state
+
+    mask = np.zeros((height, width), dtype=bool)
+    for _ in range(num_circles):
+        x = rs.randint(r_max, width - r_max)
+        y = rs.randint(r_max, height - r_max)
+        r = rs.randint(r_min, r_max)
+        spy, spx = np.ogrid[-x : width - x, -y : height - y]
+        circle = (spx * spx + spy * spy) <= r * r
+
+        mask[circle] = True
+
+    return mask
+
+
+def add_salt_and_pepper_noise(data, p=0.05):
+    if p <= 0:
+        return data
+
+    original_dtype = data.dtype
+
+    random_image_data = np.random.choice([0, 1], p=[p, 1 - p], size=data.shape)
+
+    return (data.astype(np.float32) * random_image_data).astype(original_dtype)
+
+
+def generate_label_with_noise(height, width, label_key="label", noisy_key="noisy", pred_key="pred", num_slices=1):
+    label = generate_synthetic_binary_segmentation(height, width, num_circles=10, r_min=10, r_max=50)
+    noisy_invert = ~add_salt_and_pepper_noise(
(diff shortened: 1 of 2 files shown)


tests/unit/scribbles/test_transforms.py
```diff
@@ -0,0 +1,292 @@
...
+        mask[
+            :,
+            test_input["roi"][0] : test_input["roi"][1],
+            test_input["roi"][2] : test_input["roi"][3],
+            test_input["roi"][4] : test_input["roi"][5],
+        ] = True
+        target = result["scribbles"]
+        target[~mask] = input_param["scribbles_bg_label"]
+
+        np.testing.assert_allclose(target, mask.astype(np.float32) * result["scribbles"], rtol=1e-4)
```
aihsani, line 244: isn't the target integer results? why is this comparison done in `float32` with a tolerance?

is the transform here not deterministic?

tests/unit/scribbles/test_transforms.py
```diff
@@ -0,0 +1,292 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import unittest
+
+import numpy as np
+from parameterized import parameterized
+
+from monailabel.scribbles.transforms import (
```
aihsani, line 17: These transform (along with the unit tests here) should be moved to MONAI core soon

tests/unit/scribbles/test_transforms.py
```diff
@@ -0,0 +1,292 @@
...
+    def test_add_bg_roi_transform(self, input_param, test_input, expected_shape):
+        result = AddBackgroundScribblesFromROId(**input_param)(test_input)
+        mask = result["scribbles"].astype(bool)
+        mask[
+            :,
+            test_input["roi"][0] : test_input["roi"][1],
+            test_input["roi"][2] : test_input["roi"][3],
+            test_input["roi"][4] : test_input["roi"][5],
+        ] = True
+        target = result["scribbles"]
```
aihsani, line 241: `target` should be something predetermined before the test is run and/or independent from the output of `AddBackgroundScribblesFromROId` so you can compare the `result["scribbles"]` against, not something derived from the result of `AddBackgroundScribblesFromROId`

aihsani: e.g. `target` it should come in along with `expected_shape`

aihsani (changes requested):

tests/unit/scribbles/test_transforms.py
masadcv: You are right. Target needs to be built directly from roi input and then compared against output. I will correct this.

tests/unit/scribbles/test_transforms.py
masadcv: This transform works on a "scribbles" input data, which is received from slicer as fp32 input. You are right that this could be done with integers, but it makes it easier to mix scribbles with logits and input image volume if it is fp32.

The transform, therefore, does not impose conversion of scribbles to integers or float32. It just adds values on top of what is already received (be it interger or fp32). 

Maybe there is a better way to be less ambiguous here? Perhaps two tests with 1) integers and 2) fp32 as input?

tests/unit/scribbles/test_transforms.py
masadcv: Fixed this with using original input and making the output target on the fly.

tests/unit/scribbles/test_transforms.py
masadcv: I have updated with two points:

1. Compare output directly without conversion to any data type
       `np.testing.assert_allclose(input_data, result["scribbles"], rtol=1e-4)`
2. Run the transform once with fp32 input and again with integer input. 

Let me know if this addresses your comment?

tests/unit/scribbles/test_transforms.py
```diff
@@ -0,0 +1,310 @@
...
+        np.testing.assert_equal(input_data, result["scribbles"])
+        self.assertTupleEqual(expected_shape, result["scribbles"].shape)
+        self.assertTupleEqual(test_input["scribbles"].shape, result["scribbles"].shape)
+
+    @parameterized.expand(TEST_CASE_OPTIM_TX)
+    def test_optimisation_transforms(self, input_param, test_input, output, expected_shape):
+        input_param.update({"post_proc_label": "pred"})
+        for current_tx in [ApplyGraphCutOptimisationd, ApplySimpleCRFOptimisationd]:
+            result = current_tx(**input_param)(test_input)
+            np.testing.assert_allclose(output["target"], result["pred"], rtol=1e-4)
```
aihsani, line 271: should this be considered also for a more deterministic `assert_equal`?

tests/unit/scribbles/test_transforms.py
```diff
@@ -0,0 +1,310 @@
...
+            self.assertTupleEqual(expected_shape, result["pred"].shape)
+
+    @parameterized.expand(TEST_CASE_ISEG_OPTIM_TX)
+    def test_interactive_graphcut_optimisation_transform(self, input_param, test_input, output, expected_shape):
+        input_param.update({"post_proc_label": "pred"})
+        result = ApplyISegGraphCutPostProcd(**input_param)(test_input)
+        np.testing.assert_allclose(output["target"], result["pred"], rtol=1e-4)
```
aihsani, line 278: is there a better alternative to `assert_allclose`?

tests/unit/scribbles/test_transforms.py
```diff
@@ -0,0 +1,310 @@
...
+        result = MakeISegUnaryd(**input_param)(test_input)
+
+        # make expected unary output
+        expected_result = output["target"].copy()
+        eps = np.finfo(expected_result.dtype).eps
+        expected_result[expected_result == 0] = eps
+        expected_result[expected_result == 1] = 1 - eps
+
+        # compare
+        np.testing.assert_allclose(expected_result, result["pred"], rtol=1e-4)
```
aihsani, line 293: <same a s above>

tests/unit/scribbles/test_transforms.py
```diff
@@ -0,0 +1,310 @@
...
+    @parameterized.expand(TEST_CASE_MAKE_LIKE_HIST_TX)
+    def test_make_likelihood_histogram(self, input_param, test_input, output, expected_shape):
+        input_param.update({"post_proc_label": "pred"})
+        result = MakeLikelihoodFromScribblesHistogramd(**input_param)(test_input)
+
+        # make expected output
+        expected_result = np.argmax(output["target"].copy(), axis=0)
+
+        # compare
+        np.testing.assert_allclose(expected_result, np.argmax(result["pred"], axis=0), rtol=1e-4)
```
aihsani, line 305: <same as above>

aihsani (approved):

tests/unit/scribbles/test_transforms.py
masadcv: Since these are running optimisers inside, they cannot be guaranteed to have deterministic output. I will check with assert_equal or perhaps we can lower the tolerance.

masadcv: Okay, I have switched these to equal and all seems to work for this simple examples. In future if these pose an issue, we can switch to allclose with a tolerance.

tests/unit/scribbles/test_transforms.py
masadcv: Agreed!

aihsani (approved):

masadcv:
Thanks!

PR #376, opened 2021-09-16
Add unit tests DeepEdit transforms

diazandr3s:

- Add unit tests for TestClickRatioAddRandomGuidanced transform
- Add uni test for SingleLabelSingleModalityd transform


diazandr3s:
These unit tests follow the MONAI pattern

tests/unit/deepedit/test_transforms.py
```diff
@@ -12,30 +12,124 @@
...
+    (5, 5),
+]
+
+
+class TestDiscardAddGuidanced(unittest.TestCase):
+    @parameterized.expand([DISCARD_ADD_GUIDANCE_TEST_CASE])
+    def test_correct_results(self, arguments, input_data, expected_result):
+        add_fn = DiscardAddGuidanced(arguments)
+        result = add_fn(input_data)
+        self.assertEqual(result["image"].shape, expected_result)
```
masadcv, line 108: Is it possible to verify the contents of the output against expected values?

tests/unit/deepedit/test_transforms.py
```diff
@@ -12,30 +12,124 @@
...
-    def test_t1(self):
-        DiscardAddGuidanced(keys="image")(DATA_1)
 
-    def test_t2(self):
-        ResizeGuidanceCustomd("label", "image")(DATA_1)
+class TestSingleLabelSingleModalityd(unittest.TestCase):
+    @parameterized.expand([SINGLE_LABEL_SINGLE_MODALITY_TEST_CASE_1])
+    def test_correct_results(self, arguments, input_data, expected_result):
+        result = SingleLabelSingleModalityd(**arguments)(input_data)
+        self.assertEqual(result["image"].shape, expected_result)
```
masadcv, line 132: Same here

aihsani (approved):

tests/unit/deepedit/test_transforms.py
diazandr3s: Thanks for the comment! In this case verifying the content isn't necessary as the transform only add or discard tensors in the image. So it is more important to check shape than content.

masadcv: @diazandr3s thanks! I understand, the purpose of unittests is to check for the unexpected. What if the transform unexpectedly changes values in the remaining channel? Would that be covered? Also in this case, since the last two channels are discarded, you do need to check whether that operation was performed successfully or something else happened. To me checking contents would make a lot sense here.

diazandr3s: Many thanks @masadcv 
I just added that bit in this PR (https://github.com/Project-MONAI/MONAILabel/pull/391)

PR #378, opened 2021-09-16
Remove Python3.6 support

evberrypi:
New Azure test suite runs all tests using python 3.6-3.9. On version 3.6 the integration tests do not pass.
Bumping the description for Pypi to say minimum version is Python3.7

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -16,7 +16,7 @@ project_urls =
     Source Code=https://github.com/Project-MONAI/MONAILabel
 
 [options]
-python_requires = >= 3.6
+python_requires = >= 3.7
 # for compiling and develop setup only
 # no need to specify the versions so that we could
 # compile for multiple targeted versions.


lassoan:
FYI, 3D Slicer is still on Python 3.6.7. We [plan to update](https://github.com/Slicer/Slicer/issues/5014) within a few weeks. The MONAILabel server can use a different virtual Python environment, but it can be more convenient for users that they can use Slicer's Python environment for running both the server and client.

If merging this pull request would mean that MONAILabel package on PyPI will not be updated anymore for Python-3.6 then you might consider holding up the merge until https://github.com/Slicer/Slicer/issues/5014 is resolved.

aihsani:
@lassoan MONAILabel server may run on the same machine as Slicer, but I'm not sure it matters what version it uses since all communication happens over the HTTP API. Do users use Slicer's Python environment to run the MONAI Label server currently?

lassoan:
> Do users use Slicer's Python environment to run the MONAI Label server currently?

I assume developers have Python installed and are familiar with how to set up virtual Python environments. However, end users, who just want to run a MONAILabel server might not have anything else than Slicer's Python environment.

Probably there are very few such "end users", so forcing them to install Python and learn how to manage virtual environments would not be a big deal. But if you can tolerate some failing tests for a few weeks then you could just leave the minimum Python requirement as is for the time being. If the errors are in the test suite only and not actual errors in MONAILabel, then you could even disable those problematic tests for Python-3.6 to reduce the noise.

lassoan:
I'll add a comment here when Slicer's Python is upgraded.

jcfr:
Support for Python 3.9.10 has just been integrated. Tonight preview build are expected to build against Slicer 3.9.10, we will monitor and address remaining issues tomorrow.

PR #403, opened 2021-09-21
Improve stability of histogram based likelihood, with initialisation to drichlet prior dist

masadcv:
This is to improve stability of histogram based likelihood computation for scirbbles flow. 

In cases where only foreground scribbles are given, this will prevent it to fail as it initialises conjugate prior for histogram as a special case of Dirichlet distribution. Vice versa for cases where only background scribbles are given. 


monailabel/scribbles/utils.py
```diff
@@ -110,13 +110,18 @@ def make_iseg_unary(
 
 
 def make_histograms(image, scrib, scribbles_bg_label, scribbles_fg_label, bins=32):
+    # initialise prior distribution with a special case of Dirichlet distribution
+    # this helps with the case where only foreground scribbles are given,
+    # and vice versa for background scribbles
+    prior_values = [i / bins for i in range(bins)]
```
tvercaut, line 116: Not sure I fully understand this choice of prior values. The Dirichlet prior is typically applied on the histogram directly rather than changing the values to feed the histogram. Are you assuming that the image intensities are between 0 and 1 here and this is why using a set of values spaced by 1/bins adds a pseudocount of one in each histogram bin? This would merit a comment at least.

Also, in the future it might be interesting to generalise this approach. Rather than adding a pseudo-count equal to one to each bin, it would be helpful to enable the provision of a generic pseudocount vector. This may for example be used to provide a stronger/weaker uniform prior (e.g. by setting the initial pseudo-counts to 100 if stronger, or 0.1 if weaker, everywhere) or to provide a more informative prior (e.g. by computing normalised histograms based on a previous segmentation estimate and scaling these prior histograms to give more or less weight to the prior).

monailabel/scribbles/utils.py
```diff
@@ -110,13 +110,18 @@ def make_iseg_unary(
...
+
     # collect background voxels
-    values = image[scrib == scribbles_bg_label]
+    values = np.concatenate([image[scrib == scribbles_bg_label], prior_values], axis=0)
```
tvercaut, line 119: As per the comment above rather than concatenating values it might be more intuitive to add a prior histogram of pseudo-count and the scribble-based histogram (with `density=False`) and then normalising the summed histogram.

monailabel/scribbles/utils.py
masadcv: Many thanks for looking through and pointing this out. It indeed makes more sense to do a histogram of pseudo-counts and add it to the likelihood as prior. You are right, the choice here is to add one value to each bin which is what i/bins is doing. I concatenated this to avoid Nan errors in building histogram with empty sets, but I realise I could have used `density=False` and done the other way around. 1 count was used simply to fix the error we are getting with empty sets but it does make more sense to have individual pseudo counts to generalise this better.

To make it more generalised, I will add an alpha parameter as input. This can be a scalar or a list of numbers. In case of scalar, it will be expanded to equal pseudo-counts everywhere (e.g. alpha=1 will be expanded to counts=[1, 1, 1,....]). In the other case where this is list, a check would ensure it is same shape as number of bins and it will be directly applied.

I will add a commit to address this.

monailabel/scribbles/utils.py
masadcv: Thanks, I will update the code.

monailabel/scribbles/utils.py
masadcv: Forgot to mention, the intensities are indeed assumed to be in range [0, 1] for histogram computation and are converted in the following lines:
https://github.com/Project-MONAI/MONAILabel/blob/6cb513a8f56b7eeaba996a846b884cdf9c6465d8/monailabel/scribbles/utils.py#L132

masadcv: This has been addressed now.

monailabel/scribbles/utils.py
```diff
@@ -109,22 +109,46 @@ def make_iseg_unary(
     return unary_term
 
 
-def make_histograms(image, scrib, scribbles_bg_label, scribbles_fg_label, bins=32):
+def make_histograms(image, scrib, scribbles_bg_label, scribbles_fg_label, alpha=1, bins=32):
```
tvercaut, line 112: It would make sense to offer the option to specify `alpha_bg` and `alpha_fg` independently. This is helpful if we have informative prior knowledge of these.

monailabel/scribbles/utils.py
```diff
@@ -109,22 +109,46 @@ def make_iseg_unary(
...
+    # alpha forms the psuedo-counts for Dirichlet distribution used here as
```
tvercaut, line 113: Type in "psuedo-counts"

monailabel/scribbles/utils.py
masadcv: updated to address this

monailabel/scribbles/utils.py
masadcv: fixed the typo : psuedo-counts --> pseudo-counts

diazandr3s (approved):

PR #408, opened 2021-09-22
Weekly Build Fix

aihsani:
Weekly build broken due to broken repo tree due to OHIF build.
Affects: weekly build and release
This change addresses weekly build

- Split OHIF Build into different sections of weekly build
- Commit files generated by OHIF before wheel dist


.github/workflows/weekly-preview.yml
```diff
@@ -22,15 +22,37 @@ jobs:
...
+        git commit -m "OHIF Update $YEAR_WEEK"
+        git status
+        git describe --tags --dirty --always --long
     - name: Build distribution
       run: |
-        export BUILD_OHIF=true
+        export BUILD_OHIF=false
         export HEAD_COMMIT_ID=$(git rev-parse HEAD)
         sed -i 's/name\ =\ monailabel$/name\ =\ monailabel-weekly/g' setup.cfg
         echo "__commit_id__ = \"$HEAD_COMMIT_ID\"" >> monailabel/__init__.py
```
wyli, line 54: this line trying to keep record of `HEAD_COMMIT_ID` of the release, so that when the user report issues we know exactly the version of the source code. But in this case the `HEAD_COMMIT_ID` is pointing at a workaround commit made at line 46 `git commit -m "OHIF Update $YEAR_WEEK"`

wyli (approved):
thanks @aihsani, this looks good to me, it's a very useful workaround to build with clean tags.

.github/workflows/weekly-preview.yml
aihsani: I added a section to export the HEAD commit id before any additional commits in the scripts

Issue #247, opened 2021-07-26
CI/CD - Running GPU based unit/integration tests


SachidanandAlle:
Currently GPU based test-cases are skipped.  
Need to redefine CI Task resources to use GPU.

evberrypi:
Added in #414

PR #414, opened 2021-09-23
CI/CD: Add Azure GPU pipeline

evberrypi:
This PR adds a new Azure pipeline to run GPU tests on.
It is connected to a Virtual Machine Scale Set that will provision GPU machines, run the integration tests, then spin the machines down.

This pipeline is set to never trigger, and will need to be triggered manually by an Admin, however it should be discussed if this should run on a schedule instead (eg via the weekly build) for future releases.


aihsani (approved):

aihsani:
Closes #247

PR #415, opened 2021-09-23
CI/CD: Update Azure release pipeline

evberrypi:
- Adding fixes from #411 to the Azure release pipeline
- Deleting the Github Action `release.yml` as the Azure release pipeline replaces it with this change.


.azure/azure-release-pipeline.yml
```diff
@@ -11,46 +11,80 @@ stages:
               python.version: '3.8'
           maxParallel: 1
       steps:
+      - script: |
```
aihsani, line 14: can we not name these scripts to see a summary of what each does?

aihsani (changes requested):

.azure/azure-release-pipeline.yml
evberrypi: The name goes below the script (see line 48) by convention with Azure

aihsani:
Please add TestPyPI push as a "publish" check

PR #416, opened 2021-09-23
CI/CD: Remove duplicate build pipeline

evberrypi:
With the Azure build-test pipeline sufficiently validated, we can remove the file `.github/workflows/pythonapp.yml` that served the same function, but written as a GH Action. Currently, both are running in CI.

+ Updated Azure pipeline to auto-cancel in-progress CI validation for new runs on the same PR
- Removed 'Trigger:' from Azure build pipeline -- this is being called once branches merge to main and is unnecessary.


SachidanandAlle:
We need build-docs before we can merge this PR and remove github checks.. otherwise things are stable and we can migrate to azure once we have build-docs alias available on azure

PR #473, opened 2021-10-28
Adding a basic windowing method to 3D Slicer plugin and Server App

### LINKED ISSUE #447: Setting default window width (WW) and default window level (WL)
We have had some feedback from clinical partners regarding setting of window for highlighting different organs.
The window/levels are defined typically by window width (WW) and window level (WL). This is then used in the following formula to set the min/max levels for CT images.
w_min = WL - WW/2
w_max = WL+WW/2

More info about this can be found here: https://radiopaedia.org/articles/windowing-ct?lang=gb

There are known WL/WW values for different organs/tissues (as listed in the above link). 

For example, a lung window is typically defined as WW=1500 and WL=-600. This has the following effect on visualizing the image volume:

Before applying lung window:
image

After applying lung window (WW=1500, WL=-600):
image

You can notice that it highlights tissues/lesions inside lung better. The above is just one example, the values for WW/WL can vary for different organs and institute.

In 3D Slicer, the above can be achieved as:
```
WL=-600; WW=1500
volumeNode = slicer.mrmlScene.GetFirstNodeByClass("vtkMRMLScalarVolumeNode")
volumeNode.GetDisplayNode().SetAutoWindowLevel(False)
volumeNode.GetDisplayNode().SetWindowLevelMinMax(WL-WW/2, WL+WW/2)
```

Does this sound like something we can pass as argument from an app as default, given that we know the target organ/tissue and its WL/WW. In cases where these are not present, we can skip updating the above. In other cases where the user wants to change these, there is a window level button that can be used to adjust manually. 

cc: @tvercaut

masadcv:

Fixes #447 

More info in #447 

This PR adds viewer->slicer settings panel to option tab. This includes the WW/WL for windowing in CT images.
More particularly, the following options are added:
image
The user can modify these values, which will be applied when the next sample is loaded.

On the Server App side, added a way to set WW/WL in settings sent to client.

Added the windowing settings for soft tissue to segmentation_spleen app. Here is a before and after:

Before windowing:
image

After windowing for soft tissue:
image

Let me know your comments/review, so I may improve this further. Thanks!


masadcv:
Lets discuss this in coming weeks and work towards possibly closing his PR.
I am happy to modify/add anything required or close this if not relevant for MONAI Label.

masadcv:
Closing this due to lack of activity. Let's revisit it if there is interest in getting this supported in monailabel. Thanks!

PR #486, opened 2021-11-03
Flag added for taking Epistemic mean instead of sum

finalelement:
Prior to this fix, the epistemic entropy based score was directly being used by taking the sum across the image, which gets affected by the size of the image.

The latter can play a role in selection of sample, however it is users choice to alternatively select the mean of entropy which will allow for removal of the effect of the size of the image.

Signed-off-by: vnath <vnath@nvidia.com>

### DIFF
--- a/monailabel/tasks/scoring/epistemic.py
+++ b/monailabel/tasks/scoring/epistemic.py
@@ -29,9 +29,7 @@ class EpistemicScoring(ScoringMethod):
     First version of Epistemic computation used as active learning strategy
     """
 
-    def __init__(
-        self, model, network=None, transforms=None, roi_size=(128, 128, 64), num_samples=10
-    ):
+    def __init__(self, model, network=None, transforms=None, roi_size=(128, 128, 64), num_samples=10):
         super().__init__("Compute initial score based on dropout")
         self.model = model
         self.network = network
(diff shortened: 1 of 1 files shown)


SachidanandAlle:
CI build is failing ..

SachidanandAlle:
How computing mean among entropy across multiple runs but for the same image helps?

monailabel/tasks/scoring/epistemic.py
```diff
@@ -29,14 +29,17 @@ class EpistemicScoring(ScoringMethod):
     First version of Epistemic computation used as active learning strategy
     """
 
-    def __init__(self, model, network=None, transforms=None, roi_size=(128, 128, 64), num_samples=10):
+    def __init__(
+        self, model, network=None, transforms=None, roi_size=(128, 128, 64), num_samples=10, mean_score_flag=False
+    ):
```
SachidanandAlle, line 34: We want this to enabled by default?

monailabel/tasks/scoring/epistemic.py
```diff
@@ -29,7 +29,9 @@ class EpistemicScoring(ScoringMethod):
     First version of Epistemic computation used as active learning strategy
     """
 
-    def __init__(self, model, network=None, transforms=None, roi_size=(128, 128, 64), num_samples=10):
+    def __init__(
+        self, model, network=None, transforms=None, roi_size=(128, 128, 64), num_samples=10, mean_score_flag=False
```
SachidanandAlle, line 33: , mean_score_flag=False is not needed anymore

SachidanandAlle (approved):

finalelement:
@SachidanandAlle hit merge, its blocked at my end.

PR #487, opened 2021-11-04
Update transforms for multilabel DeepEdit

diazandr3s:

- Modify transforms to simulate clicks for all organs at the same time
- Consider background as another label - Change the concept of background clicks to just another label
- Update the **_AddGuidanceFromPointsCustomd_** to receive landmarks according to what the UI sends

To do:

- Modify the UI to send clicks corresponding to all labels


SachidanandAlle:
How about deepgrow?

SachidanandAlle (approved):

PR #528, opened 2021-11-16
Revert "enable coverage upload only for PRs in github workflow"

SachidanandAlle:
Reverts Project-MONAI/MONAILabel#527

In case if my assumption is wrong.. let @evberrypi confirm why the coverage report is not getting pushed from azure to https://app.codecov.io/gh/Project-MONAI/MONAILabel


evberrypi (approved):
This makes sense to remove, however my guess is that the actual coverage report is coming from .github/codecov.yml

PR #530, opened 2021-11-16
croppad deepedit transforms

pritesh-mehta:
I have added two transforms, which are an alternative to ResizeGuidanceCustomd, for DeepEdit pipelines that choose to crop/pad images to a certain size for CNN processing, rather than downsampling/upsampling.

Use case: Prostate segmentation. In my application, I crop the input image as pre-processing to remove background tissues, so need to adjust the guidance in line with crop.


diazandr3s (approved):

SachidanandAlle (changes requested):
How about the impact? Is it going to impact current App?
How about unit tests if it's a new code?

diazandr3s:
> How about the impact? Is it going to impact current App? How about unit tests if it's a new code?

These are very good points. 
These transforms will potentially increase clicks interaction for the deepedit.

I agree we should also include unit tests for these transforms. Happy to help with that :)

pritesh-mehta:
Unit tests added for new transforms. Please review.

SachidanandAlle (approved):

PR #551, opened 2021-12-07
Create DeepEdit using Preds and Simulated Clicks as Input

diazandr3s:

The main goal of this PR is to improve user interaction. This includes changing the transforms to also consider predictions generated by the model during the training process. 

- Fix the transforms to work with a list and not a list of lists for guidance
- Add predictions to the Input Tensor for Multilabel DeepEdit for better interaction
- Change the name to some transforms to improve readability
- Modify Tensorboard handler to visualize preds during training
- Add preds tensor to values equal to 1/# of labels in Interaction class for the case when NO Interaction is performed. Specifically changed these lines: https://github.com/Project-MONAI/MONAILabel/blob/main/monailabel/deepedit/multilabel/interaction.py#L92. 
- Send the mask name from UI to the server. This should be sent along with the clicks! 
- Modify transform **_AddGuidanceSignalAndPredsd_** to add received predictions from UI
- Made it work for missing labels as done here: https://github.com/Project-MONAI/MONAILabel/blob/main/monailabel/deepedit/multilabel/transforms.py#L826

The training and testing are being tested. The only bits missing are:

- **Properly send predictions (predicted label) from Slicer UI.** Currently, UI sends the image name to be read from the datastore (labels/original). This doesn't consider potential changes made by the user with manual tools in the client.
- **What's the best way to preprocess the predictions (predicted label) sent by UI**, using the same transforms for the "image". Should we always pass the predictions through pre transforms?

Commets: 

- **Interaction has improved when the user provides clicks on areas that are not covered by the predictions.** 
- Even though the labels are correctly generated using softmax, **Slicer UI shows two labels in the same voxels** when modifying the labels using clicks. 

schema_deepedit_multilabel_preds


SachidanandAlle:
Closing this PR.. @diazandr3s you might want to consider it differently based on new app structure

PR #561, opened 2021-12-19
[WIP] Add multilabel DeepEdit App for komp dataset

diazandr3s:

- Testing multilabel deepedit on komp dataset. Mice dataset

### DIFF
--- a/sample-apps/deepedit_multilabel_komp/README.md
+++ b/sample-apps/deepedit_multilabel_komp/README.md
@@ -1,0 +1,62 @@
+# DeepEdit Multilabel Default App
+
+### Model Overview
+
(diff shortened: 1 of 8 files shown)


SachidanandAlle:
Instead of of creating a new app.. think of single app that can work on both datasets.. if transforms/pretrained model are different.. you can add/use --conf option

diazandr3s:
> Instead of of creating a new app.. think of single app that can work on both datasets.. if transforms/pretrained model are different.. you can add/use --conf option

You're right. I'll remove this PR once I discuss the results with Murat :)

diazandr3s:
Closing PR. 
The only/main difference with the sample multilabel deepedit is the dictionary containing label names and numbers: https://github.com/Project-MONAI/MONAILabel/pull/561/files#diff-8d40f00d4cd8d22a6dae835529baab7ce70d0b85acbe13a0cbefad7ac4b415d0R43

PR #567, opened 2021-12-28
Add property to Slicer plugin to automatically open segment editor after next sample was fetched

jlvahldiek:
Small enhancement; can save the user one annoying click.

- Added MONAILabel property MONAILabel/autoOpenSegmentEditor to control if user wants to automatically open segment editor after next sample was fetched
- Added logic to initSample() to activate Slicer's segment editor automatically after next sample was fetched

### DIFF
--- a/plugins/slicer/MONAILabel/MONAILabel.py
+++ b/plugins/slicer/MONAILabel/MONAILabel.py
@@ -124,18 +124,7 @@ class _ui_MONAILabelSettingsPanel(object):
             "valueAsInt",
             str(qt.SIGNAL("valueAsIntChanged(int)")),
         )
-        
-        autoOpenSegmentEditorCheckBox = qt.QCheckBox()
-        autoOpenSegmentEditorCheckBox.checked = False
-        autoOpenSegmentEditorCheckBox.toolTip = "Enable this option to automatically open segment editor after Next Sample was fetched"
-        groupLayout.addRow("Auto-Open Segment Editor:", autoOpenSegmentEditorCheckBox)
-        parent.registerProperty(
-            "MONAILabel/autoOpenSegmentEditor",
-            ctk.ctkBooleanMapper(autoOpenSegmentEditorCheckBox, "checked", str(qt.SIGNAL("toggled(bool)"))),
-            "valueAsInt",
-            str(qt.SIGNAL("valueAsIntChanged(int)")),
-        )
-        
+
         developerModeCheckBox = qt.QCheckBox()
         developerModeCheckBox.checked = False
         developerModeCheckBox.toolTip = "Enable this option to find options tab etc..."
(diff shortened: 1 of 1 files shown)


SachidanandAlle:
Can you fix the static checks...

Easiest way is.. 
./runtests.sh --codeformat

SachidanandAlle:
https://github.com/Project-MONAI/MONAILabel/blob/main/CONTRIBUTING.md

jlvahldiek:
I am trying and trying.. An old black version on my local machine is the problem. Will fix it and re-submit.

SachidanandAlle:
Thank you.. 😊

SachidanandAlle:
I am confused here.. why we need to open segment edit after next sample.. the objective of MONAI Label is to promote AI features/models for users... like DeepEdit/Deepgrow, Segmentation Models etc..  This option contradicts the basics objective of monai label (AI driven)

jlvahldiek:
I can understand that. We use MONAILabel for various segmentation tasks. One of them is 2D segmentation of X-ray images from scratch with the goal to create a corresponding model for pre-segmentations, which in turn will serve for HITL tasks.
One of our human annotators came up with the suggestion to simplify the workflow by opening the SegmentEditor directly after clicking NextSample in the simple human segmentation task described above.
All of our human annotators welcomed this simplification. I thought it might be of interest to some of the community as well, so I raised the PR. If it is not interesting, no problem. We can always use our own fork of the Slicer plugin.

Thank you very much.

SachidanandAlle:
This is fine if it's helping some community.. 
default option is disabled.. so not a problem :)

SachidanandAlle (approved):

lassoan:
It is also an option to add a [qMRMLSegmentEditorWidget](http://apidocs.slicer.org/master/classqMRMLSegmentEditorWidget.html) to the MONAILabel module GUI (for example, in a collapsible section). Then it is not necessary to switch between modules but you can make quick manual corrections without leaving the MONAILabel module. Having the qMRMLSegmentEditorWidget in the module GUI also allows simplifications, such as removal of node segmentation node selectors (the segmentation node is set automatically) and limit the list of effects (e.g., you can choose to display only manual tools such as paint/draw/erase/scissors).

jlvahldiek:
> It is also an option to add a [qMRMLSegmentEditorWidget](http://apidocs.slicer.org/master/classqMRMLSegmentEditorWidget.html) to the MONAILabel module GUI (for example, in a collapsible section). Then it is not necessary to switch between modules but you can make quick manual corrections without leaving the MONAILabel module. Having the qMRMLSegmentEditorWidget in the module GUI also allows simplifications, such as removal of node segmentation node selectors (the segmentation node is set automatically) and limit the list of effects (e.g., you can choose to display only manual tools such as paint/draw/erase/scissors).

Sounds interesting. Will look into it. Thanks.

SachidanandAlle:
For now.. i have added this change.. meanwhile you can continue to explore the qMRMLSegmentEditorWidget 
Thanks for your contribution 👍

jlvahldiek:
I was able to embed the qMRMLSegmentEditorWidget into the MONAILabel extension, as suggested by @lassoan in a collapsible section:

Bildschirmfoto 2022-01-01 um 22 46 53

In my opinion it is a neat enhancement to the Slicer extension. What do you think, @SachidanandAlle? Where should it be placed? I would prefer after "active learning" section.

SachidanandAlle:
This should be fine after active learning section.. Looks good to me.
However by default I suggest to keep it in collapsed mode..

Issue #566, opened 2021-12-24
Incorrect filename truncation


SlicerMorph:
**Describe the bug**

When I use the Next Sample button, the retrieved displayed volume name is incorrect. That's because i believe parser incorrectly cuts at the volume name at "-" sign of the volume. Below is the log from the python interactor, displaying the correct volume name. In slicer, regardless of the volume, all retrieved volumes are displayed as __15.5_baseline__ in Slicer, which confuses the user. 

```
Check if file exists/shared locally: /workspace/undeterminedSex_AAMN_N305-17-e15.5_baseline.nii.gz => False
http://127.0.0.1:8000/datastore/image?image=undeterminedSex_AAMN_N305-17-e15.5_baseline
```
![image](https://user-images.githubusercontent.com/45026482/147365618-c92572ef-0904-4509-ae7a-c2da8cdf01bb.png)

SachidanandAlle:
It's not a bug.. the last 20 chars of original filename is shown to avoid combo box stretched by too much.. and becomes difficult to read/identify

SlicerMorph:
It may not be a programmatic bug, it is a bug for user. How am I going to know which file I am segmenting then?

lassoan:
Full string could be set as tooltip.

SachidanandAlle:
Currently logs can help you to get full details.. u can see them under python console inside slicer...
Or as suggested by @lassoan .. if you can add a small fix to add tooltip and raise a PR.. will be helpful

pwrightkcl:
I need a way to view the full filename in Slicer without access to the logs. I will set up and run the server, but a clinician will label the images on a different PC, so using the server log is impractical for my use case (the clinician would need to note the time he loaded an image, and I would need to go back and track it down, which creates extra work and risk of error).

My images have necessarily long filenames. When I load an image in Slicer manually, I can see the full filename in the Volumes module. The display window truncates the filename, but whereas MONAI label just shows the last few characters, Slicer truncates as "head...tail", which is more likely to be identifiable (in my case most of my filename tails are identical).

Would it be possible to have MONAI label use its truncated filename only for its dropdown, but let Slicer handle the full filename the normal way?

pwrightkcl:
I don't know how to do that, sorry.

diazandr3s:
Thanks for reporting this, @pwrightkcl.

[...]

PR #568, opened 2021-12-28
Fix for truncated filenames in Slicer extension

jlvahldiek:
Fix for #566.

Added a tooltip with the full file name to the inputSelector of the Slicer extension in order to clarify file name truncation of long file names.

### DIFF
--- a/plugins/slicer/MONAILabel/MONAILabel.py
+++ b/plugins/slicer/MONAILabel/MONAILabel.py
@@ -426,7 +426,6 @@ class MONAILabelWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
         self.ui.inputSelector.clear()
         for v in self._volumeNodes:
             self.ui.inputSelector.addItem(v.GetName())
-            # for truncated long filenames show tooltip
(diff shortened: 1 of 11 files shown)


plugins/slicer/MONAILabel/MONAILabel.py
```diff
@@ -413,6 +413,7 @@ def updateGUIFromParameterNode(self, caller=None, event=None):
         self.ui.inputSelector.clear()
         for v in self._volumeNodes:
             self.ui.inputSelector.addItem(v.GetName())
+            self.ui.inputSelector.setToolTip(self.current_sample["name"])
```
SachidanandAlle, line 416: will be better if we have condition check here..
`self.current_sample.get("name", "") if self.current_sample else ""`

SachidanandAlle: there can be a case where next sample is not fetched.. and user can load his/her own image in slicer and continue to work as temp session... in that case.. i believe current_sample will be empty.. so i think it's better to have this check.. as we allow users to bring his/her own images

jlvahldiek: changed accordingly :-)

SachidanandAlle (approved):

Issue #565, opened 2021-12-23
Overlapping segmentations are flattened, information is lost


jlvahldiek:
If two segments are drawn overlapping in Slicer ('Allow overlap'), the segmentation file generated by the MONAILabel plugin is flattened and information is lost.

Initial segmentation with two overlapping segments:
<img width="1405" alt="Bildschirmfoto 2021-12-23 um 15 35 26" src="https://user-images.githubusercontent.com/583023/147254997-4647e70d-d671-45fb-8ad3-d4162c39a9d2.png">

After "Submit label" via MONAILabel plugin and manual load of image/segmentation file (NIFTI or NRRD) into Slicer:
<img width="1408" alt="Bildschirmfoto 2021-12-23 um 15 36 49" src="https://user-images.githubusercontent.com/583023/147255487-83ef2bd8-d589-4f8f-9e7a-1a69fbf5e781.png">

A warning message about possible loss of overlapping information in not displayed.

Possible solution/workaround I am using at the moment:
1. change file extension of MONAILabel plugin to '.seg.nrrd'
2. change onSaveLabel() in plugins/slicer/MONAILabel/MONAILabel.py to save directly from the segmentation node instead of flattening the segments by saving from a binary labelmap node (lines 1204-1228)

Do you propose any other solutions or workflows? Should I raise a PR?

SachidanandAlle:
It is an open source.. your contributions will be helpful.. please raise a pr to support non overlapping segmentation..  you might want to add a setting for the user to control it..

SachidanandAlle:
Thanks for adding this feature

diazandr3s:
It'll be helpful to know in which use cases overlapping segments are used/interesting? When training a multilabel App, we use a one-hot representation of the labels, which means non-overlapping labels ...

jlvahldiek:
We use MONAILabel for different tasks. One of them is to annotate a large set of 2D Xray images from scratch by several human annotators (radiologists). Anatomic structures as well as pathological findings can overlap each other in Xray images. That's why we raised PR #579.

At the moment we are very happy with this solution. Feedback from the annotators regarding annotation workflow was also excellent (compared to other segmentation software we used before).

At the moment we are working on anatomic pre-segmentations for a human-in-the-loop task also on 2D Xrays with MONAILabel. In this task there will also be overlapping segments.

diazandr3s:
Superb! Thanks for your reply, @jlvahldiek

PR #578, opened 2022-01-04
Support overlapping segments in Slicer extension

jlvahldiek:
Fix for issue #565.

- introduce MONAILabel setting "allowOverlappingSegments"
  - default status is False
  - if set to True and MONAILabel file extension is ".seg.nrrd" segments are directly saved from the segmentation node which prevents flattening of overlapping segments/information loss
  - if set to False segments are saved from a labelmap node (unchanged prior implementation)


jlvahldiek:
messed up, will re-submit

PR #579, opened 2022-01-04
Support overlapping segments in Slicer extension

jlvahldiek:
Fix for issue #565.

- introduce MONAILabel setting "allowOverlappingSegments"
  - default status is False
  - if set to True and MONAILabel file extension is ".seg.nrrd" segments are directly saved from the segmentation node which prevents flattening of overlapping segments/information loss
  - if set to False segments are saved from a labelmap node (unchanged prior implementation)

### DIFF
--- a/plugins/slicer/MONAILabel/MONAILabel.py
+++ b/plugins/slicer/MONAILabel/MONAILabel.py
@@ -1202,10 +1202,10 @@ class MONAILabelWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
         try:
             qt.QApplication.setOverrideCursor(qt.Qt.WaitCursor)
             segmentationNode = self._segmentNode
-            labelmapVolumeNode = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLLabelMapVolumeNode")
-            slicer.modules.segmentations.logic().ExportVisibleSegmentsToLabelmapNode(
-                segmentationNode, labelmapVolumeNode, self._volumeNode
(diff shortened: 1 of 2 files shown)


plugins/slicer/MONAILabel/MONAILabel.py
```diff
@@ -1258,7 +1282,13 @@ def onSaveLabel(self):
             label_in = tempfile.NamedTemporaryFile(suffix=self.file_ext, dir=self.tmpdir).name
             self.reportProgress(5)
 
-            slicer.util.saveNode(labelmapVolumeNode, label_in)
+            if (
+                slicer.util.settingsValue("MONAILabel/allowOverlappingSegments", True, converter=slicer.util.toBool)
+                and slicer.util.settingsValue("MONAILabel/fileExtension", self.file_ext) == ".seg.nrrd"
```
SachidanandAlle, line 1287: `.seg.nrrd`? why we need to hard-fix this extension?

jlvahldiek: 1) .seg.nrrd makes sure all layers are preserved
2) .seg.nrrd makes sure that everything is exactly reproduced when label file is loaded back into Slicer

please see also discussion at: https://discourse.slicer.org/t/how-to-correctly-export-overlapping-segmentations/21050/

pynrrd and slicerio are useful python packages to read .seg.nrrd files

SachidanandAlle (approved):

Issue #605, opened 2022-02-03
denseCRF3D import error


rijobro:
In a clean conda environment, I did the following to install MONAILabel (wanted it from source so I could use my branch)

```bash
conda create -n monai-label python=3.9 -y
conda activate monai-label
conda install pytorch cudatoolkit=11.3 -c pytorch -y
cd <path/to/MONAILabel>
pip install -e .
```
Then the following line causes an error
```bash
python -c "monailabel start_server ..."
```

And more specifically, so does the following
```bash
python -c "from monailabel.scribbles.transforms import denseCRF3D"
```

Both times the error is as follows:

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/rich/Documents/Code/MONAILabel/monailabel/scribbles/transforms.py", line 16, in <module>
    import denseCRF3D
ImportError: /home/rich/miniconda3/envs/monai-label/lib/python3.9/site-packages/denseCRF3D.cpython-39-x86_64-linux-gnu.so: undefined symbol: _ZSt28__throw_bad_array_new_lengthv
```

Any idea what's going on?

SachidanandAlle:
@masadcv can you help and unblock @rijobro

SachidanandAlle:
until @masadcv responds.. you can disable the scribbles 

For example:
https://github.com/Project-MONAI/MONAILabel/blob/main/sample-apps/deepedit/main.py#L27
https://github.com/Project-MONAI/MONAILabel/blob/main/sample-apps/deepedit/main.py#L143-L146

SachidanandAlle:
@masadcv can you handle the import error and automatically exclude scribbles in the app if import fails due to xyz reasons..

diazandr3s:
Thanks, @rijobro for reporting this. 
I agree with @SachidanandAlle. Please comment those lines out while this issue is solved.
There is another conversation around simpleCRF [here](https://github.com/HiLab-git/SimpleCRF/issues/7#issuecomment-1027264172). 
Let's see how this progress :)

masadcv:
Thanks @SachidanandAlle  and @diazandr3s !
I am looking into this.

@rijobro this error is coming from SImpleCRF installation - I am unsure why it is causing this. As Andres mentioned, there is another ticket to resolve some installation issues with SimpleCRF on windows. 

[...]

PR #606, opened 2022-02-03
Quick fix for simplecrf imports

masadcv:

Work around for #605 until the error is debugged.

### DIFF
--- a/monailabel/scribbles/transforms.py
+++ b/monailabel/scribbles/transforms.py
@@ -30,8 +30,8 @@ from .utils import (
     maxflow3d,
 )
 
-densecrf, _ = optional_import("dennseCRF")
-densecrf3d, _ = optional_import("denseCRF3D")
+densecrf, has_densecrf = optional_import("denseCRF")
+densecrf3d, has_densecrf3d = optional_import("denseCRF3D")
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


masadcv:
@SachidanandAlle I made all SImpleCRF imports optional. Please give it a try by uninstalling SimpleCRF and running any app. On my end this exp runs fine.

monailabel/scribbles/transforms.py
```diff
@@ -31,6 +30,10 @@
     maxflow3d,
 )
 
+densecrf, _ = optional_import("dennseCRF")
+densecrf3d, _ = optional_import("denseCRF3D")
```
rijobro, line 34: here, you could use:
```suggestion
densecrf3d, has_densecrf3d = optional_import("denseCRF3D")
```
and then you could use `has_densecrf3d` elsewhere to display an error if the user tries to use scribbles.

masadcv: thanks! will update this.

rijobro: Looks good!

SachidanandAlle (approved):

SachidanandAlle:
This solves one part of the problem... when SimpleCRF is installed by import fails due to xyz reasons...
I guess there is another thread where installation itself can be simplified with pre-built binaries while installing SimpleCRF package

Issue #609, opened 2022-02-04
PersistentDataset effectively not used


rijobro:
User can choose to use `PersistentDataset`, but `cache_dir` is passed as `None`, which from the documentation: 

> If `cache_dir` is `None`, there is effectively no caching.

I feel that the default location should be in the app folder.

PR #610, opened 2022-02-04
Add persistent dataset dir

rijobro:
Fixes https://github.com/Project-MONAI/MONAILabel/issues/609.

By default the persistent dataset dir is adjacent to the model checkpoint (typically at the base level of the sample app):

```
./sample-apps/deepedit_multilabel_VerSe/PersistentDir
```

### DIFF
--- a/monailabel/tasks/train/basic_train.py
+++ b/monailabel/tasks/train/basic_train.py
@@ -69,13 +69,14 @@ class Context:
         self.val_batch_size = None  # validation batch size
         self.device = None  # device on which training will run
         self.network = None  # network
-        self.dataset_type = "CacheDataset"  # dataset type
+        self.dataset_type = "PersistentDataset"  # dataset type
         self.dataloader_type = "ThreadDataLoader"  # dataloader type
         self.pretrained = False  # using pretrained model
         self.max_epochs = 1  # max epochs to run training
         self.multi_gpu = False  # multi gpu enabled
         self.local_rank = 0  # local rank in case of multi gpu
         self.world_size = 0  # world size in case of multi gpu
+        self.persistent_dir = None  # dir of PersistentDataset (if used)
 
         self.request = None
         self.trainer = None
@@ -107,6 +108,7 @@ class BasicTrainTask(TrainTask):
         model_dict_key="model",
         find_unused_parameters=False,
         load_strict=False,
+        persistent_dir=None,
     ):
         """
         :param model_dir: Base Model Dir to save the model checkpoints, events etc...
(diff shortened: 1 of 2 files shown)


SachidanandAlle:
Doesn't seem a right/standard option.. lets think of a better solution for this

rijobro:
No, but `app_dir` isn't available in the constructor, whereas `model_dir` is. To pass `app_dir` would presumably involve updating all the sample apps, too.

SachidanandAlle:
Actually this is little weird.. MONAI class should have give default implementation rather than None.. which is no useful for any user.. if I/user have to pass None, i/user would rather use simple Dataset.. why I shall be using Persistent Dataset with cache_dir as none... God knows!

However to have a fix... I would rather suggest something like this..

https://github.com/Project-MONAI/MONAILabel/blob/main/monailabel/tasks/train/basic_train.py#L188

```
        ds_dir = os.path.join(pathlib.Path.home(), ".cache", "monailabel", "dataset")
        os.makedirs(ds_dir, exist_ok=True)
        cache_dir = tempfile.mkdtemp(dir=ds_dir) # or any other way to random generate the temp file name and reduce this to single line of code..
        shutil.rmtree(cache_dir) # we need name only

        dataset = (
            CacheDataset(datalist, transforms)
            if context.dataset_type == "CacheDataset"
            else SmartCacheDataset(datalist, transforms, replace_rate)
            if context.dataset_type == "SmartCacheDataset"
            else PersistentDataset(datalist, transforms, cache_dir=cache_dir)
            if context.dataset_type == "PersistentDataset"
            else Dataset(datalist, transforms)
        )

```

rijobro:
It's obviously up to you guys, but I would err away from saving the cache in `$HOME/.cache`, as this will quickly get messy. Especially if you have different instances all using the same data source. If they have different transforms, but the filenames are the same, you'll end up with erroneous results.

SachidanandAlle:
You r creating a temp folder for every run.. 

cache_dir = tempfile.mkdtemp(dir=ds_dir)

SachidanandAlle:
now with this kind of change.. `model/{output_folder}/cache_{run_id}` will have all the temp cache items and it gets removed on successful train finishes

SachidanandAlle (approved):

SachidanandAlle:
Thanks for identifying and debugging this issue :)

rijobro:
Thanks for the updates. I think the naming doesn't make sense to save the cached dataset in a folder called `model`. 

Also, by using `run_id`, you lose the ability to re-use the persistent cache between different runs. This is safer (modifying transforms invalidates the cache) but slower (identical runs could have re-used the cache).

PR #633, opened 2022-02-13
Allow SSL options as args to run server in https mode

SachidanandAlle:


diazandr3s (approved):
This is great! I know several users will like this new security feature

PR #671, opened 2022-03-01
Fix docs build dependency

SachidanandAlle:

because of python3.7 may be latest 1.22.x numpy is not available )
https://readthedocs.org/projects/monailabel/builds/16224068/


wyli (approved):
Thanks, it looks good to me.

PR #693, opened 2022-03-15
Show/hide tabs based on models that are available

SachidanandAlle:

### DIFF
--- a/plugins/slicer/MONAILabel/MONAILabel.py
+++ b/plugins/slicer/MONAILabel/MONAILabel.py
@@ -479,10 +479,26 @@ class MONAILabelWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
 
         if self.models and [k for k, v in self.models.items() if v["type"] == "segmentation"]:
             self.ui.segmentationCollapsibleButton.collapsed = False
+            self.ui.segmentationCollapsibleButton.show()
+        else:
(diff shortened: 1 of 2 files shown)


diazandr3s (approved):

plugins/slicer/MONAILabel/MONAILabel.py
```diff
@@ -479,10 +479,26 @@ def updateGUIFromParameterNode(self, caller=None, event=None):
...
         if self.models and [k for k, v in self.models.items() if v["type"] == "scribbles"]:
             self.ui.scribblesCollapsibleButton.collapsed = False
+            self.ui.scribblesCollapsibleButton.show()
+        else:
+            self.ui.scribblesCollapsibleButton.hide()
+
+        if self.info.get("trainers", {}):
+            self.ui.aclCollapsibleButton.show()
+        else:
+            self.ui.aclCollapsibleButton.hide()
```
jlvahldiek, line 501: @SachidanandAlle, we are running two segmentation apps that do not make use of any trainers.

With these code changes our workflow of using MONAILabel to segment images from scratch or to use pre-segmentations does not work any more because the buttons 'Next Sample' and 'Submit Sample' are not available any more.

I can understand the rationale behind these code changes but for our use case it doesn't work. Is it possible to discuss these code changes? Thanks.

PR #707, opened 2022-03-26
Add spatial size as iput for infer class - UNETR configuration

diazandr3s:

- Add argument spatial size in infer class
- Trained and uploaded the UNETR for single label model

### DIFF
--- a/sample-apps/radiology/main.py
+++ b/sample-apps/radiology/main.py
@@ -219,6 +219,7 @@ def main():
     conf = {
         "models": "deepedit",
         "use_pretrained_model": "false",
+        "network": "unetr",
     }
 
     app = MyApp(app_dir, studies, conf)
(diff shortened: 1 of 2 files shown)


sample-apps/radiology/main.py
```diff
@@ -219,6 +219,7 @@ def main():
     conf = {
         "models": "deepedit",
         "use_pretrained_model": "false",
+        "network": "unetr",
```
SachidanandAlle, line 222: The model is huge.. making it default is not a good idea..
And these config will be still part of config/deepedit..

Refer config/segmentation on how to add config per model

SachidanandAlle (changes requested):

sample-apps/radiology/main.py
SachidanandAlle: I guess this part of main function.. example to run inference.. in that case.. its fine..

SachidanandAlle (approved):

sample-apps/radiology/main.py
diazandr3s: Thanks for your comments ... I'll change this bit

Issue #607, opened 2022-02-04
Scribbles only contains foreground/background


rijobro:
I'm working on multilabel segmentation. 

All of the labels are displayed in the deepgrow panel:

![image](https://user-images.githubusercontent.com/33289025/152505272-5052436c-29b0-44a6-a809-a96a86ce7757.png)

But only foreground and background are displayed in the Scribbles panel:

![image](https://user-images.githubusercontent.com/33289025/152504782-6722be28-508c-461d-b24d-960ee3d8e9b5.png)

Is there any way to do multilabel scribbles?

SachidanandAlle:
@masadcv - Like SmartEdit, you can list the labels.. allow the single operation/scribbles at a time.. the result can be updated over selected label.. User can repeat the operation N times without changing the label selection.. this way user can annotate one label at a time.. without much code changes.. 

In later stages, you can allow to save the previous foreground/background when user selects a different label..

masadcv:
@SachidanandAlle thanks for the input. I will look into this in coming days. One major difference of this from click-based methods is that you can have a full segmentation volume with multiple labels, which may be easier to do multilabel as you will just end up adding interactions to the segmentation volume.

On the server side, we need to update transforms to handle multilabel. Currently, employed Histogram+GraphCut may not be insufficient for multilabel problems - we may need to replace GraphCut with CRF.

masadcv:
The latest version (current main) now contains a solution to multi-label scribbles.
See: #717 and related #731

@rijobro can you please test scribbles now with your use-case - let me know if you spot any issues with multilabel scribbles.

I am marking this as resolved for now

rijobro:
@masadcv thanks for this! I can't test as I'm currently doing research that doesn't use MONAI Label, but will feedback if there are any issues when I'm next using it.

PR #716, opened 2022-03-31
Fix labels from Slicer plugin offset by extra background class

adamaji:
Signed-off-by: Adam Aji <3487395+adamaji@users.noreply.github.com>

This PR addresses https://github.com/Project-MONAI/MONAILabel/discussions/713. When running the Slicer plugin, I saw that having the "background" segment visible would affect scribbles by offsetting the expected labels of "2"/"3" for "background_scribbles"/"foreground_scribbles" to "3"/"4". This can also affect saving labels via "Submit Label" by incrementing all classes after "background" by 1.

I'm assuming this offset isn't intended from other references in the code
https://github.com/Project-MONAI/MONAILabel/blob/f9e37feeb7a27d39e924a7a61c4ec9e50be9cdc0/monailabel/scribbles/infer.py#L103-L104

The main change here is to replace `ExportVisibleSegmentsToLabelmapNode` with `ExportSegmentsToLabelmapNode` and specify exactly which segments should be exported. For scribbles, I opted to export only the selected `currentSegment` with the foreground and background scribble segments.

Because of that, this is a potential soft-workaround for https://github.com/Project-MONAI/MONAILabel/issues/607, as it should allow you to associate the scribbles with different segments in the segmentation. It's still somewhat unclear that that `currentSegment` combobox is connected to scribbles since it lives under the "SmartEdit / DeepGrow" heading, but since scribbles was already using it, I thought this should be fine for now.

### DIFF
--- a/plugins/slicer/MONAILabel/MONAILabel.py
+++ b/plugins/slicer/MONAILabel/MONAILabel.py
@@ -1279,8 +1279,9 @@ class MONAILabelWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
             qt.QApplication.setOverrideCursor(qt.Qt.WaitCursor)
             segmentationNode = self._segmentNode
             labelmapVolumeNode = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLLabelMapVolumeNode")
-            slicer.modules.segmentations.logic().ExportVisibleSegmentsToLabelmapNode(
-                segmentationNode, labelmapVolumeNode, self._volumeNode
+            save_segment_ids = [segment_id for segment_id in segmentationNode.GetSegmentation().GetSegmentIDs() if segment_id != "background"]
+            slicer.modules.segmentations.logic().ExportSegmentsToLabelmapNode(
+                segmentationNode, save_segment_ids, labelmapVolumeNode, self._volumeNode
             )
 
             segmentation = segmentationNode.GetSegmentation()
@@ -1722,8 +1723,11 @@ class MONAILabelWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
             # get scribbles + label
             segmentationNode = self._segmentNode
             labelmapVolumeNode = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLLabelMapVolumeNode")
-            slicer.modules.segmentations.logic().ExportVisibleSegmentsToLabelmapNode(
-                segmentationNode, labelmapVolumeNode, self._volumeNode
+            _, segment = self.currentSegment()
+            selected_label_name = segment.GetName()
+            save_segment_ids = [selected_label_name, "background_scribbles", "foreground_scribbles"]
+            slicer.modules.segmentations.logic().ExportSegmentsToLabelmapNode(
+                segmentationNode, save_segment_ids, labelmapVolumeNode, self._volumeNode
             )
             scribbles_in = tempfile.NamedTemporaryFile(suffix=self.file_ext, dir=self.tmpdir).name
             self.reportProgress(5)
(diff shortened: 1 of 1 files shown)


plugins/slicer/MONAILabel/MONAILabel.py
```diff
@@ -1279,8 +1279,9 @@ def onSaveLabel(self):
             qt.QApplication.setOverrideCursor(qt.Qt.WaitCursor)
             segmentationNode = self._segmentNode
             labelmapVolumeNode = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLLabelMapVolumeNode")
-            slicer.modules.segmentations.logic().ExportVisibleSegmentsToLabelmapNode(
-                segmentationNode, labelmapVolumeNode, self._volumeNode
+            save_segment_ids = [segment_id for segment_id in segmentationNode.GetSegmentation().GetSegmentIDs() if segment_id != "background"]
```
SachidanandAlle, line 1282: yes, we can discard the `background` label (which is normally used by deepedit multilabel model) along with scribbles background/foreground when you save the label..
but the fix possibly by adding "background" to the list mentioned at: https://github.com/Project-MONAI/MONAILabel/blob/main/plugins/slicer/MONAILabel/MONAILabel.py#L1293

plugins/slicer/MONAILabel/MONAILabel.py
```diff
@@ -1722,8 +1723,11 @@ def onUpdateScribbles(self):
             # get scribbles + label
             segmentationNode = self._segmentNode
             labelmapVolumeNode = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLLabelMapVolumeNode")
-            slicer.modules.segmentations.logic().ExportVisibleSegmentsToLabelmapNode(
-                segmentationNode, labelmapVolumeNode, self._volumeNode
+            _, segment = self.currentSegment()
+            selected_label_name = segment.GetName()
+            save_segment_ids = [selected_label_name, "background_scribbles", "foreground_scribbles"]
+            slicer.modules.segmentations.logic().ExportSegmentsToLabelmapNode(
```
SachidanandAlle, line 1729: there is a known issue for scribbles.. it currently can't support/persist scribbles per label.. specially in case for multi-label.. what you need is save/persist the bg/fg scribbles (similar to fg/bg fudical/click points) per label.. so the behavior/user experience in case of multilabel might not be good.

However, i will let @masadcv to decide what he needs from the label node.. i guess he might want to get all labels and pick fg/bg label masks as part of pre-transforms + `current label name` (i guess he needs current label name in case of he wants all labels; which might be the reason, it's not working) and send back the results with corrections..

again only corrected label should get updated.. i guess it needs to be fixed/verified as well..

In future these methods might want to use existing label masks as input along with fg/bg label masks..
Or until that future comes, it's ok for me to include only fg/bg scribbles sent to server to run the scribbles action..

plugins/slicer/MONAILabel/MONAILabel.py
```diff
@@ -1753,9 +1757,7 @@ def onUpdateScribbles(self):
 
             # display result from server
             self.reportProgress(90)
-            _, segment = self.currentSegment()
-            label = segment.GetName()
-            self.updateSegmentationMask(result_file, [label])
+            self.updateSegmentationMask(result_file, [selected_label_name])
```
SachidanandAlle, line 1760: yes,, this will make sure we update the selected mask only 👍

plugins/slicer/MONAILabel/MONAILabel.py
masadcv: @adamaji @SachidanandAlle understood.
I would like to keep labels and do selection on the server side. This is to make it future-proof where we can use more advanced multi-label post-processing methods such as CRF that take into account other labels as well.

The scribbles labels are currently hard-coded in server app, these need to be passed from the client - in case the order changes or similar shift in label are present (due to any changes to client). Hence, a need for dynamic updates for this. 

As Sachi said, background may need to be removed before saving final segmentation output, but this can be done at https://github.com/Project-MONAI/MONAILabel/blob/main/plugins/slicer/MONAILabel/MONAILabel.py#L1293

I will continue this in PR #717 where I will merge the changes from here for updating the current label and provide a generalisable solution to the problem that we can extend in future to multi-label problems. As this needs proper testing with multilabel problem, I will take sometime and get back with an update when ready.

adamaji:
Closing because this PR will be superseded by work from #717 .

PR #717, opened 2022-04-01
[WIP] Implement multilabel as binary scribbles for a given segment and add support for dynamic scribbles label ids from client/server

masadcv:
As reported in this discussion #713 and workaround attempted in #716 the labels for scribbles are fixed which causes issues when these change in client side. In addition, as reported in the discussion that "update" from scribbles functions relies on selected label in SmartEdit tab. This need not be selected or provided explicitly for scribbles functionality to work.

To address the above, I have the following changes in this PR
- [x] Send `label_info` for any update scribbles call in Slicer plugin
- [x] Recieve and update from `labels` from scribbles server app in Slicer plugin
- [x] Read and use scribbles label indices from `label_info` in Server app
- [x] Define default labels in scribbles Server app to avoid any confusion/ambiguity when running apps with Client
- [x] Add combobox for Scribbles labels selection
- [x] Add mechanism to reset scribbles for change in label selection with a use warning
- [x] Add relevant updates from #716 (thanks adamaji!)
- [x] Test with multi-label selected
- [x] Code format
- [x] Unittests locally
- [x] Tested locally with scribbles examples

### DIFF
--- a/monailabel/scribbles/transforms.py
+++ b/monailabel/scribbles/transforms.py
@@ -76,6 +76,27 @@ class InteractiveSegmentationTransform(Transform):
 
         return d
 
+    def _set_scribbles_idx_from_labelinfo(self, d):
+        def convert_string(in_val):
+            if isinstance(in_val, str):
+                # OHIF format - given as "1+idx" string - extract idx
+                return int(in_val.split("+")[-1])
+            else:
+                # slicer format - already int
+                return in_val
+
+        label_info = d.get("label_info", [])
+        for lb in label_info:
+            if lb.get("name", None) == "background_scribbles":
+                id = convert_string(lb.get("id", self.scribbles_bg_label))
+                self.scribbles_bg_label = id
+                logging.info("Loading background scribbles labels from: {} with index: {}".format(lb.get("name"), id))
+
+            if lb.get("name", None) == "foreground_scribbles":
+                id = convert_string(lb.get("id", self.scribbles_fg_label))
+                self.scribbles_fg_label = id
+                logging.info("Loading foreground scribbles labels from: {} with index: {}".format(lb.get("name"), id))
+
 
 #######################################
 #######################################
@@ -100,6 +121,9 @@ class AddBackgroundScribblesFromROId(InteractiveSegmentationTransform):
 
     def __call__(self, data):
         d = dict(data)
+
+        # load scribbles idx from labels_info (if available)
+        self._set_scribbles_idx_from_labelinfo(d)
 
         # read relevant terms from data
         scribbles = self._fetch_data(d, self.scribbles)
@@ -170,6 +194,9 @@ class MakeLikelihoodFromScribblesHistogramd(InteractiveSegmentationTransform):
 
     def __call__(self, data):
         d = dict(data)
+
+        # load scribbles idx from labels_info (if available)
+        self._set_scribbles_idx_from_labelinfo(d)
 
         # copy affine meta data from image input
         d = self._copy_affine(d, src=self.image, dst=self.post_proc_label)
@@ -290,6 +317,9 @@ class MakeISegUnaryd(InteractiveSegmentationTransform):
     def __call__(self, data):
         d = dict(data)
 
+        # load scribbles idx from labels_info (if available)
+        self._set_scribbles_idx_from_labelinfo(d)
+
         # copy affine meta data from image input
         self._copy_affine(d, self.image, self.unary)
 
@@ -383,6 +413,9 @@ class ApplyISegGraphCutPostProcd(InteractiveSegmentationTransform):
         # attempt to fetch algorithmic parameters from app if present
         self.lamda = d.get("lamda", self.lamda)
         self.sigma = d.get("sigma", self.sigma)
+
+        # load scribbles idx from labels_info (if available)
+        self._set_scribbles_idx_from_labelinfo(d)
 
         # copy affine meta data from image input
         d = self._copy_affine(d, src=self.image, dst=self.post_proc_label)
--- a/monailabel/scribbles/transforms.py
+++ b/monailabel/scribbles/transforms.py
@@ -77,23 +77,15 @@ class InteractiveSegmentationTransform(Transform):
         return d
 
     def _set_scribbles_idx_from_labelinfo(self, d):
-        def convert_string(in_val):
-            if isinstance(in_val, str):
-                # OHIF format - given as "1+idx" string - extract idx
-                return int(in_val.split("+")[-1])
-            else:
-                # slicer format - already int
-                return in_val
-
         label_info = d.get("label_info", [])
         for lb in label_info:
             if lb.get("name", None) == "background_scribbles":
-                id = convert_string(lb.get("id", self.scribbles_bg_label))
+                id = lb.get("id", self.scribbles_bg_label)
                 self.scribbles_bg_label = id
                 logging.info("Loading background scribbles labels from: {} with index: {}".format(lb.get("name"), id))
 
             if lb.get("name", None) == "foreground_scribbles":
-                id = convert_string(lb.get("id", self.scribbles_fg_label))
+                id = lb.get("id", self.scribbles_fg_label)
                 self.scribbles_fg_label = id
                 logging.info("Loading foreground scribbles labels from: {} with index: {}".format(lb.get("name"), id))
 
--- a/plugins/slicer/MONAILabel/MONAILabel.py
+++ b/plugins/slicer/MONAILabel/MONAILabel.py
@@ -219,7 +219,7 @@ class MONAILabelWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
 
         self.scribblesMode = None
         self.ignoreScribblesLabelChangeEvent = False
-        self.multi_label = False
+        self.deepedit_multi_label = False
 
     def setup(self):
         """
@@ -630,8 +630,11 @@ class MONAILabelWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
         self.ui.dgPositiveControlPointPlacementWidget.setEnabled(self.ui.deepgrowModelSelector.currentText)
         self.ui.dgNegativeControlPointPlacementWidget.setEnabled(self.ui.deepgrowModelSelector.currentText)
 
-        self.multi_label = "background" in self.info.get("labels", [])
-        if self.multi_label:
+        self.deepedit_multi_label = False
+        m = self.models.get(self.ui.deepgrowModelSelector.currentText) if self.models else None
+        self.deepedit_multi_label = m and m.get("type") == "deepedit" and len(m.get("labels")) > 0
+
+        if self.deepedit_multi_label:
             self.ui.dgLabelBackground.hide()
             self.ui.dgNegativeControlPointPlacementWidget.hide()
             self.ui.freezeUpdateCheckBox.show()
@@ -876,7 +879,7 @@ class MONAILabelWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
         for i in range(n):
             coord = pointListNode.GetNthControlPointPosition(i)
 
-            world = [0, 0, 0, 0]
+            world = [0, 0, 0]
             pointListNode.GetNthControlPointPositionWorld(i, world)
 
             p_Ras = [coord[0], coord[1], coord[2], 1.0]
@@ -896,7 +899,7 @@ class MONAILabelWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
 
         coord = pointListNode.GetNthControlPointPosition(index)
 
-        world = [0, 0, 0, 0]
+        world = [0, 0, 0]
         pointListNode.GetNthControlPointPositionWorld(index, world)
 
         p_Ras = [coord[0], coord[1], coord[2], 1.0]
@@ -1490,7 +1493,7 @@ class MONAILabelWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
             qt.QApplication.setOverrideCursor(qt.Qt.WaitCursor)
 
             sliceIndex = None
-            if self.multi_label:
+            if self.deepedit_multi_label:
                 params = {}
                 segmentation = self._segmentNode.GetSegmentation()
                 for name in self.info.get("labels", []):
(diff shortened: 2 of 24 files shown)


adamaji:
Hi, thanks for the help here. Trying this out with the same sample app setup from #713 I'm not seeing scribbles work still. It seems like `ExportVisibleSegmentsToLabelmapNode` is still being used here, so the labels are still offset, both for scribbles and for "Submit Label". Also, is scribbles supposed to work when given a multilabel input? From https://github.com/Project-MONAI/MONAILabel/issues/607 I thought that scribbles is currently intended to be a single-class algorithm.

Would it be possible to work off of #716 instead, and to address the issue of 

>may only work for this specific case, given the user has selected required label in SmartEdit tab.

we could implement the first part of @SachidanandAlle 's suggestion here https://github.com/Project-MONAI/MONAILabel/issues/607#issuecomment-1030450954 -- which I understood to be adding a similar label combobox in the scribbles section to select the desired label to perform graphcut on?

masadcv:
Can you try this with a fresh setup? Please remember to also update your slicer plugin to the one in this PR. 
It should address any labels offsets, so no need to change export functions. 

Multilabel is a separate problem from this, which isn't used in the app you are testing. Let's solve one problem at a time.

adamaji:
Another reason I'm saying to use `ExportSegmentsToLabelmapNode` instead of `ExportVisibleSegmentsToLabelmapNode` is because the latter function makes the labels sent to the server dependent on the visibility of the segments in the MRML scene. So if you make any of the segments hidden with the "eye" button or through code, they don't get sent over. I don't think that is intended behavior, and if it is, it's unclear from the GUI that it is. `ExportSegmentsToLabelmapNode` will allow you to specify which segments specifically get sent, so you don't have to do any segment deletion either.

Additionally, I think multilabel is something that should be addressed to some degree here. I think future proofing by sending all the necessary labels is fine, but is scribbles just not supposed to work currently if you have more than 1 label class? It seems like to make things clear to the user here, we need to connect the `selected_label_name` from `self.updateSegmentationMask(result_file, [selected_label_name])` directly to the scribbles section of the plugin. This could be an additional combobox in the scribbles section.

SachidanandAlle:
`ExportSegmentsToLabelmapNode` is fine 👍

SachidanandAlle:
possibly @adamaji and @masadcv can sync up together and put the best pieces together to make the scribbles better...
thank you both :)

masadcv:
> Additionally, I think multilabel is something that should be addressed to some degree here. I think future proofing by sending all the necessary labels is fine, but is scribbles just not supposed to work currently if you have more than 1 label class? It seems like to make things clear to the user here, we need to connect the `selected_label_name` from `self.updateSegmentationMask(result_file, [selected_label_name])` directly to the scribbles section of the plugin. This could be an additional combobox in the scribbles section.

Many thanks for your input and for discussion. Here we have multiple things going on, and hence I mentioned multilabel to be a separate problem. The current scribbles infer task only looks at binary problems (i.e. using graphcuts). I believe to properly address this we may want to use a different transform (e.g. CRF) that handles multilabels together. This could be an additional PR itself, so best to not touch that here. 

I agree with the suggestion that we can use `selected_label_names` in a combobox specifically for scribbles to divide this problem into multiple binary segmentation problems - instead of multilabel together. Infact I already have some update on this, will push code in a moment (it is still WIP though)

I am off now, but would be good to catchup with you next week. Please send me a message on slack (if you are there)...

adamaji:
Sounds good @masadcv , I'll ping you on slack next week

masadcv:
@SachidanandAlle This is ready for your review. 
I have added `ExportSegmentsToLabelmapNode` and syncing with @adamaji to see if there is anything left from his initial PR.

monailabel/scribbles/transforms.py
```diff
@@ -76,6 +76,27 @@ def _copy_affine(self, d, src, dst):
 
         return d
 
+    def _set_scribbles_idx_from_labelinfo(self, d):
+        def convert_string(in_val):
+            if isinstance(in_val, str):
+                # OHIF format - given as "1+idx" string - extract idx
```
SachidanandAlle, line 82: this should go into client side.. server should have only one standard format.. anything specific let it happen in the client side.. and send the required label info

plugins/slicer/MONAILabel/MONAILabel.py
```diff
@@ -1280,26 +1340,30 @@ def onSaveLabel(self):
...
+            save_segment_ids = vtk.vtkStringArray()
             for idx, segmentId in enumerate(segmentIds):
                 segment = segmentation.GetSegment(segmentId)
-                if segment.GetName() in ["foreground_scribbles", "background_scribbles"]:
+                if segment.GetName() in ["background", "foreground_scribbles", "background_scribbles"]:
                     logging.info(f"Removing segment {segmentId}: {segment.GetName()}")
                     segmentationNode.RemoveSegment(segmentId)
                     continue
 
+                save_segment_ids.InsertNextValue(segmentId)
```
adamaji, line 1357: Potentially, segment removal can be dropped now that `save_segment_ids` is defined? I'm not sure that the segment removal is necessary for any further downstream tasks.

```suggestion
                if segment.GetName() not in ["background", "foreground_scribbles", "background_scribbles"]:
                    save_segment_ids.InsertNextValue(segmentId)
```

plugins/slicer/MONAILabel/MONAILabel.py
```diff
@@ -1724,9 +1788,20 @@ def onUpdateScribbles(self):
             # get scribbles + label
             segmentationNode = self._segmentNode
             labelmapVolumeNode = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLLabelMapVolumeNode")
-            slicer.modules.segmentations.logic().ExportVisibleSegmentsToLabelmapNode(
-                segmentationNode, labelmapVolumeNode, self._volumeNode
+            save_segment_ids = vtk.vtkStringArray()
+            segmentationNode.GetSegmentation().GetSegmentIDs(save_segment_ids)
```
adamaji, line 1792: is the "background" label useful for the scribbles / crf algorithm? If not, I might suggest removing that segment ID here

adamaji:
Gave this a run with the sample setup described in https://github.com/Project-MONAI/MONAILabel/discussions/713#discussion-3978347 and saw that it works for that case. I haven't used any of the OHIF stuff so can't confirm there. Dropped some comments -- but sending the addition label->index mapping info makes sense to me, especially considering future proofing for further multi label enhancements

plugins/slicer/MONAILabel/MONAILabel.py
masadcv: Current binary graphcut method may not need it. 
Background will be required for multi-label CRF-based methods (see my last message on slack regarding multi-label). 
I will therefore keep this and handle any label selection on server side...

plugins/slicer/MONAILabel/MONAILabel.py
masadcv: Agreed! I am removing `RemoeSegment` in latest commit.

monailabel/scribbles/transforms.py
masadcv: alright! will update this to a single standard format (int) on server and conversions on client-side.

plugins/slicer/MONAILabel/MONAILabel.py
```diff
@@ -524,6 +526,35 @@ def updateGUIFromParameterNode(self, caller=None, event=None):
         idx = 0 if idx < 0 < self.ui.labelComboBox.count else idx
         self.ui.labelComboBox.setCurrentIndex(idx)
 
+        if self._segmentNode:
```
SachidanandAlle, line 529: this logic can be combined with above logic where we populate items for self.ui.labelComboBox

plugins/slicer/MONAILabel/MONAILabel.py
```diff
@@ -524,6 +526,35 @@ def updateGUIFromParameterNode(self, caller=None, event=None):
...
+            segmentation = self._segmentNode.GetSegmentation()
+            totalSegments = segmentation.GetNumberOfSegments()
+            segmentIds = [
+                segmentation.GetNthSegmentID(i)
+                for i in range(totalSegments)
+                if segmentation.GetNthSegmentID(i) not in ["background", "background_scribbles", "foreground_scribbles"]
+            ]
+            currentItems = [self.ui.scribLabelComboBox.itemText(i) for i in range(self.ui.scribLabelComboBox.count)]
+            if len(segmentIds) == len(currentItems) and all([a == b for a, b in zip(segmentIds, currentItems)]):
```
SachidanandAlle, line 538: i think we don't need this.. we always populate the combo box and try to select previously selected label using parameterNode

plugins/slicer/MONAILabel/MONAILabel.py
```diff
@@ -1828,15 +1904,31 @@ def onClearScribblesSegmentNodes(self):
...
+
     def onClearScribbles(self):
+        # for clearing scribbles and resetting tools to default
+        # remove "scribbles" segments from label
+        self.onClearScribblesSegmentNodes()
+
+        self.ui.paintScribblesButton.setChecked(True)
+        self.ui.eraseScribblesButton.setChecked(False)
+
+        self.ui.scribblesSelector.setCurrentIndex(0)
```
SachidanandAlle, line 1928: how about ROI.. will it be good idea to use a specific named ROI for scribbles.. similar to foreground/background scribbles.. instead of any ROI node..

masadcv:
The OHIF scribbles seems to be broken now with this PR.  I am trying to debug it but as I have less experience with javascript, it may take a while. If someone can help me with it that would be great, otherwise I will get back to fixing it within this PR soon.

plugins/slicer/MONAILabel/MONAILabel.py
masadcv: The main reason I took this approach is because we want to show a dialogue box to user whenever a different combobox label is selected for scribbles. We want to warn and ask for ok that the existing scribbles will be cleared. 
With the above logic for self.ui.labelComboBox the combobox was updated multiple times (clear, repopulate, select current text) - this isnt good for scribbles as then we will show multiple unnecessary dialogue boxes to user. 

Here I only update combobox for scribbles if there is a change in labels, otherwise I dont touch it.

SachidanandAlle: i meant something like this.. little less duplicate code.. 

```python
self.ui.labelComboBox.clear()
self.ui.scribLabelComboBox.clear()
if self._segmentNode:
    segmentation = self._segmentNode.GetSegmentation()
    totalSegments = segmentation.GetNumberOfSegments()
    segmentIds = [segmentation.GetNthSegmentID(i) for i in range(totalSegments)]
    for idx, segmentId in enumerate(segmentIds):
        segment = segmentation.GetSegment(segmentId)
        label = segment.GetName()
        if label not in ["foreground_scribbles", "background_scribbles"]:
            self.ui.labelComboBox.addItem(label)
        if label not in ["background", "foreground_scribbles", "background_scribbles"]:
            self.ui.scribLabelComboBox.addItem(label)
else:
    for label in self.info.get("labels", {}):
        self.ui.labelComboBox.addItem(label)
        if label != "background":
            self.ui.scribLabelComboBox.addItem(label)

currentLabel = self._parameterNode.GetParameter("CurrentLabel")
idx = self.ui.labelComboBox.findText(currentLabel) if currentLabel else 0
idx = 0 if idx < 0 < self.ui.labelComboBox.count else idx
self.ui.labelComboBox.setCurrentIndex(idx)

currentScribbleLabel = self._parameterNode.GetParameter("CurrentScribbleLabel")
idx = self.ui.scribLabelComboBox.findText(currentScribbleLabel) if currentScribbleLabel else 0
idx = 0 if idx < 0 < self.ui.scribLabelComboBox.count else idx
self.ui.scribLabelComboBox.setCurrentIndex(idx)

```

masadcv: I was trying to avoid having to change combobox label and call `onSelectScribLabel()` multiple times. This code makes changes to it multiple times, each time calling the method which might not be needed if labels in these comboboxes are not changing. 
For now, I have bypassed the notification message in `onSelectScribLabel()` by wrapping this piece of code within `self.ignoreScribblesLabelChangeEvent = True`. I am adding a commit with these changes and marking this as resolved.

plugins/slicer/MONAILabel/MONAILabel.py
masadcv: addressed in latest commit.

plugins/slicer/MONAILabel/MONAILabel.py
masadcv: I need to think a bit about this one. 
There could be many ways in which we utilise ROIs. 

Currently, they are used to limit the "working area" for scribbles, as such anything outside is ignored (e.g. by adding background scribbles outside or any other similar methods on server) - so the current flow does not require them to be specific for background/foreground etc.

masadcv: We certainly need to reset ROIs for each new label selection in scribbles - I will look into this. 
I think we can also add an ROI button within scribbles panel to help a user quickly add ROI.

plugins/slicer/MONAILabel/MONAILabel.py
```diff
@@ -18,6 +18,7 @@
 import time
 import traceback
 from collections import OrderedDict
+from re import S
```
jamesobutler, line 21: Is this an intentional new import? I couldn’t find where this was now being used.

masadcv: oops, this looks like a typo - let me correct it! Thank you for pointing this out.

masadcv:
@SachidanandAlle whats the best approach here, now that we have missing scribbles functionality? 
Shall we mark this as **not to be merged PR**?

plugins/slicer/MONAILabel/MONAILabel.py
masadcv: Fixed it.

jamesobutler:
@masadcv Isn’t the removal of SimpleCRF functionality (“These are used in scribbles as post-processing steps to refine segmentations”), just an element of Scribbles, where Scribbles can continue to exist and be improved upon independently of those post-processing steps?

Can we focus effort to solve the offset label issue as proposed in #716?

masadcv:
@jamesobutler Yes, you are right these are separate things. However, as you can see there are merge conflicts now due to removal of server functionality. Therefore, there is no current method that can be used to test this within monai label, hence I want to avoid merging this for now (until we have at least one method to test against).

SachidanandAlle:
> @SachidanandAlle whats the best approach here, now that we have missing scribbles functionality? Shall we mark this as **not to be merged PR**?

yes.. marked it as WIP for now..

jamesobutler:
Should #716 be reopened and that incremental work continue separately?

SachidanandAlle:
makes sense.. to track it.. because some part of the problem still needs to be solved irrespective of method (simplecrf vs others) used for scribbles..

PR #720, opened 2022-04-02
remove icon filepath from Slicer UI file

jamesobutler:
This removes some icon filepath specifications from the UI file for the MonaiLabel module for Slicer. These paths were not actually working due to incorrect relative path to the icon. They would have been needed to be updated to something like `../Icons/refresh-icon.png`. 

I'm removing these specifications from the UI file because they are instead setup in the python code for the MonaiLabel module as seen below:
https://github.com/Project-MONAI/MONAILabel/blob/bbe26982a9b53c67d09be099034b937afcfb6769/plugins/slicer/MONAILabel/MONAILabel.py#L252-L259

### DIFF
--- a/plugins/slicer/MONAILabel/Resources/UI/MONAILabel.ui
+++ b/plugins/slicer/MONAILabel/Resources/UI/MONAILabel.ui
@@ -42,10 +42,6 @@
        <property name="text">
         <string/>
        </property>
-       <property name="icon">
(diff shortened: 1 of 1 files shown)


SachidanandAlle:
Thanks for looking into this.. and contributing to make it better 👍

SachidanandAlle (approved):

PR #721, opened 2022-04-02
Remove usage of deprecated Slicer Markups API 

jamesobutler:
This updates MonaiLabel for Slicer to not use Slicer deprecated markups API. Usage of the deprecated markups API is logged at runtime and provides the following suggestions as replacements. 

I have also updated the terminology used in the MonaiLabel module to match the desired usage in Slicer. For example instead of a "Fiducial" markups node, it is now referred to as a "Point List" node. Slicer is in the process of updating the class name to be used as vtkMRMLMarkupsPointListNode instead of vtkMRMLMarkupsFiducialNode, but that is not yet complete. 
There are various "Markups" nodes including a "Point List" node. All "Markups" nodes contain "Control Points".

It has been updated in the UI in button texts and hover help tips such as the following:
image


jamesobutler:
Since MonaiLabel is not available for the current Slicer stable (4.11.20210226) there should be no concern for backwards compatibility issues. Typically latest version of Slicer extensions are only supported for latest versions of Slicer Preview. This should work for latest Slicer Preview markups API in preparation for Slicer 5.0 to be released soon.

SachidanandAlle:
I will verify these changes and let you know by Mon/Tue

Issue #719, opened 2022-04-02
Unsuccessful `pip install monailabel>=0.4*` on Windows platform


jamesobutler:
Hi, I'm a frequent contributor to the [3D Slicer](https://github.com/Slicer/Slicer) project and was attempting to try out MonaiLabel for Slicer. We are prepping to cut a new stable release of Slicer (5.0) here soon and are checking the status of various extensions prior to release. When I attempted to set up the dependencies instructed at https://github.com/Project-MONAI/MONAILabel#development-release so that I could then test the MonaiLabel module in Slicer, I came across an install error of `monailabel`.

On the Windows platform, I'm unable to `pip install monailabel>=0.4*` successfully. See the following output below. It fails to build the [`SimpleCRF`](https://github.com/HiLab-git/SimpleCRF) whl due to missing python development headers. Slicer specifically does not include python development headers as can be reviewed by [this post](https://github.com/Slicer/Slicer/pull/6117#issuecomment-1042479854) from another Slicer core developer. This lack of a Windows whl for `SimpleCRF` appears to have been brought up previously in https://github.com/Project-MONAI/MONAILabel/issues/363 and is still a work in progress at https://github.com/HiLab-git/SimpleCRF/issues/7. Therefore getting `SimpleCRF` with a Windows whl on PyPI or reducing the number of dependencies that `monailabel` requires in order to cutout the need for `SimpleCRF` should be completed.

To help make MonaiLabel successful for the 3D Slicer community it will be necessary to make sure the install process is successful and as simple as possible for the Windows platform. Slicer version 4 recently passed the 1 million download mark where 70% of downloads were on the Windows platform (https://download.slicer.org/download-stats/).

**Describe the bug**
```powershell
PS C:\Users\JamesButler\AppData\Local\NA-MIC\Slicer 4.13.0-2022-03-31\bin> ./PythonSlicer.exe -m pip install monailabel>=0.4*
Collecting monailabel>=0.4*
  Using cached monailabel-0.4.0rc1-202203230043-py3-none-any.whl (5.1 MB)
Collecting Shapely==1.8.0
  Using cached Shapely-1.8.0-cp39-cp39-win_amd64.whl (1.1 MB)
Collecting fastapi==0.73.0
  Using cached fastapi-0.73.0-py3-none-any.whl (52 kB)
Collecting expiringdict==1.2.1
  Using cached expiringdict-1.2.1-py3-none-any.whl
Collecting openslide-python==1.1.2
[...]

PR #725, opened 2022-04-05
SimpleCRF install fix

andreped:
As mentioned in [this thread](https://github.com/HiLab-git/SimpleCRF/issues/7#issuecomment-1088632474), a fix for installing SimpleCRF that uses precompiled binaries was developed. A temporary PyPI project was made, which means that installing SimpleCRF on all platforms is much easier, and solves the issue mentioned in the issue #719. The PyPI project is called [SimpleCRF-binaries](https://pypi.org/project/SimpleCRF-binaries/).

I have made the necessary modifications to the install procedure of MONAI Label to use the new package.

However, run some unit tests just to verify that the package installs and works as expected with MONAI Label.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,7 +17,7 @@ pynetdicom==2.0.1
 expiringdict==1.2.1
 schedule==1.1.0
 timeloop==1.0.2
-SimpleCRF==0.2.1.1
+SimpleCRF-binaries==0.2.1.2
 numpy>=1.21.5
 einops>=0.3.2
 openslide-python==1.1.2
--- a/setup.cfg
+++ b/setup.cfg
@@ -43,7 +43,7 @@ install_requires =
     expiringdict==1.2.1
     schedule==1.1.0
     timeloop==1.0.2
-    SimpleCRF==0.2.1.1
+    SimpleCRF-binaries==0.2.1.2
     numpy>=1.21.5
     einops>=0.3.2
     openslide-python==1.1.2


SachidanandAlle:
Thanks for taking care of this.. however i just tried uninstall simple crf and installed new binaries.. i see some issue.. the required python modules like `maxflow`, `denseCRF`, `denseCRF3D` are not available

andreped:
@SachidanandAlle It is strange, because I have run tests to assess this and they all were successful:
https://github.com/andreped/SimpleCRF/actions/runs/2096086307

The unit tests I did can be found in this workflow: https://github.com/andreped/SimpleCRF/blob/binaries/.github/workflows/CI-build.yaml#L44

Note that I am working in the binaries branch to produce these wheels.

SachidanandAlle:
> @SachidanandAlle It is strange, because I have run tests to assess this and they all were successful: https://github.com/andreped/SimpleCRF/actions/runs/2096086307
> 
> The unit tests I did can be found in this workflow: https://github.com/andreped/SimpleCRF/blob/binaries/.github/workflows/CI-build.yaml#L44
> 
> Note that I am working in the binaries branch to produce these wheels.
> 
> I will do some testing on a local macos 10.13 desktop to see if I can reproduce the issue.

the issue was numpy again... need latest numpy.. and i couldn't use 1.21.5

SachidanandAlle:
https://github.com/Project-MONAI/MONAILabel/issues/726

andreped:
Due to the licensing issue with SimpleCRF, this PR is not longer needed, as SimpleCRF will be removed from MONAI Label.

Closing for now.

Issue #22, opened 2021-04-20
Wrong Orientation/Origin for multichannel volumes (4D) using Slicer - Issue visualizing BRATS images on 3DSlicer


SachidanandAlle:
> BRATS and prostate, which have more than one modality. It seems we're having the same problem I had before. Once Slicer loaded the volume, I didn't find a way to get back all the modalities
> 
> This was something I left in the to-solve list. As workaround solution, I get the modalities before I loaded in Slicer
> 
> In addition to the multimodality issue in Slicer, the BRATS data seems to have a problem with the orientation

 -  From @diazandr3s

SachidanandAlle:
Moving to backlog.. not a priority for 0.2 version release

lassoan:
> However, visualizing 4D images in 3D Slicer is not supported.

We can visualize 4D volumes, but loading of 4D Nifti files is currently not supported. You can load 4D NRRDs. We could add loading of 4D Nifiti images, which should not be a lot of work if ITK already supports it.

diazandr3s:
> [image.nii.gz](https://github.com/Project-MONAI/MONAILabel/files/8681272/image.nii.gz)
> 
> qform code is set to NIFTI_XFORM_UNKNOWN.
> sform code is set to NIFTI_XFORM_ALIGNED_ANAT.
> 
> According to the [specification 1](https://nifti.nimh.nih.gov/nifti-1/documentation/nifti1fields/nifti1fields_pages/qsform.html): This method is not recommended, and is present mainly for compatibility with ANALYZE 7.5 files.
> 
> It might not be recommended because its interpretation is ambiguous.
> 
> Slicer relies on ITK to read this image and it ends up being interpreted as the image defined in the LPS coordinate system, that’s why the IJKtoRAS matrix is diag(-1,-1,1,1). Maybe this is not correct, but I’m not sure if there is anybody who can tell or if there is consensus on how to interpret this image.
> 
> [label.nii.gz](https://github.com/Project-MONAI/MONAILabel/files/8681274/label.nii.gz)
> 
> Both qform and sform code are set to NIFTI_XFORM_SCANNER_ANAT.
> 
[...]

PR #729, opened 2022-04-06
[WIP] Brain application using multimodality input - BRATS

diazandr3s:

- Segmentation App for brain tumour segmentation using 4D BRATS dataset
- Please ADD the slicer plugin path to 3DSlicer instead of installing the MONAI Label module from the Extension Manager


SachidanandAlle:
Add a new model like segmentation_brain.. similar to segmentation_spleen

plugins/slicer/MONAILabel/MONAILabel.py
```diff
@@ -1195,7 +1195,7 @@ def onNextSampleButton(self):
 
             logging.info(f"Check if file exists/shared locally: {image_file} => {local_exists}")
             if local_exists:
-                self._volumeNode = slicer.util.loadVolume(image_file)
+                self._volumeNode = slicer.util.loadVolume(image_file, {"discardOrientation": True})
```
diazandr3s, line 1198: https://github.com/Project-MONAI/MONAILabel/issues/22

lassoan: Discarding the orientation is an ugly workaround, which might be temporarily acceptable for a specific data set, while working on the proper solution. Still, add plenty of comments explaining why it was necessary for this specific case. 

Discarding orientation may sound like discarding some uninteresting detail but it actually completely ruins your data set, because you can never recover from it. You generate a new data set that is spatially not related to the input data set anymore. It is so bad because you don't know what information you've discard, so you cannot ever reverse the operation.

Applying an explicit RAS->LPS conversion (by multiplying the IJKToRAS transform by `diag(-1,-1,1,1)`) is a much more manageable workaround. But if we talk about workarounds then a much cleaner solution would be use a loader to split the volume into 3D volumes before loading or use a different nifti loader to get the 4D numpy array and IJKToRAS matrix and create a volume sequence from it.

We could add 4D Nifti support to Slicer, but we need strong driving use cases. If we start seeing that 4D Nifti is adopted by the community then it would give motivation. But in this case it seems that somebody just dumped completely unrelated 3D images into a 4D sequence. I don't think this is a good practice. Image sequences are intended for storing tightly related 3D volumes (e.g., time sequence, vector components, etc.), but not for storing images that are acquired separately, using different imaging protocols (in this case, T1, T1CE, T2, FLAIR). You cannot even specify the metadata in the Nifti file (e.g., which frame stores which protocol), so this is definitely not an archival format. A user can temporarily merge these volume into a single file during training but this is certainly not necessary, and it should not be done if it causes more problems than the inconveniences of using 4 separate files.

Who created these 4D Nifti files?

lassoan: That said, if you say that 4D Nifti support is the most important missing feature in Slicer for MONAILabel then we can certainly take care of it. Just want to make sure we are working on things that add the most value.

SachidanandAlle:
Do we need this PR anymore? if not can we close this..

diazandr3s:
> Do we need this PR anymore? if not can we close this..

I'm still using this branch quite actively. Do you think it is better to keep this just as a branch only?

plugins/slicer/MONAILabel/MONAILabel.py
pieper: I agree with @lassoan 's assessment here.  Reading 4D nifti as a sequence or a vector volume would make sense if there are common enough use case that make the conventions are standardized and meaningful.  The current Slicer behavior here is suboptimal though.  I'm thinking we may need a module that imports nifti more like the way [RawImageGuess](https://discourse.slicer.org/t/new-extension-rawimageguess-for-loading-of-images-from-unrecognized-file-format/9219) deals with semi-structured data.

PR #736, opened 2022-04-10
Linting with pre-commit ci

jamesobutler:
This transitions some of the python style checks to be run using pre-commit workflows like what is done in the Monai project.

The following has been added to the contributing doc for instructions how to run the pre-commit checks locally.

>     install `pre-commit` if you have not already.
>     ```
>     python -m pip install pre-commit
>     ```
>     Run `pre-commit` check from MonaiLabel working directory.
>     ```
>     cd MonaiLabel
>     python -m pre-commit run --all-files
>     ```

This is to be paired with GitHub application pre-commit ci https://github.com/marketplace/pre-commit-ci which is a continuous integration service for the pre-commit framework. https://github.com/pre-commit/action is the deprecated GitHub actions version. Pre-commit ci is used by the Monai project and includes auto updating of PRs with fixes. This can be turned off in the configuration if desired. The pre-commit ci application once installed for this repository would have a status check like:
image


jamesobutler:
A request for this PR would be to integrate all of the commits in this PR (rebase and merge) to maintain the separate actions in their own organized commits rather than squashed down into one large commit which would make it difficult when reviewing the commit history to find out which diff change corresponds to which set of changes. This specific request is based on merged commit https://github.com/Project-MONAI/MONAILabel/commit/da07e42696bc2240549344c463d5990abf817354 which has dropped torch 1.5 support, which is not reflective in the commit title and is a separate set of changes compared to updating `PY_REQUIRED_MINOR` to version 3.7. The original PR #738 for that had those changes in 2 separate commits.

Additionally, @SachidanandAlle are you able to add the https://github.com/marketplace/pre-commit-ci application so that we can see its status check pass here in this PR? That is part of the changes in `consolidate lint checks to cross platform pre-commit framework` specifically.

SachidanandAlle:
good to sync up with monai repo on this.. it's been a while i didn't follow all the github related changes added into monai.. the same can be borrowed into monailabel.  thanks for helping on these additions...

I suggest we can add the `pre-commit` dependency also as part of [requirements-dev.txt](https://github.com/Project-MONAI/MONAILabel/blob/main/requirements-dev.txt)

And then possibly user can run `pre-commit run --all-files` as you suggested.

Also can we add mypy into the list.. ?  I find mypy is useful among the static checks..
Since it's org project, i will ask forum to to add pre-commit-ci application to enabled for the repository..

.github/workflows/docker.yml
```diff
@@ -12,11 +12,11 @@ jobs:
...
     steps:
-      - uses: actions/checkout@v2
-      - name: Set up Python 3.8
-        uses: actions/setup-python@v2
+      - uses: actions/checkout@v3
+      - name: Set up Python
+        uses: actions/setup-python@v3
         with:
-          python-version: 3.8
+          python-version: 3.9
```
SachidanandAlle, line 19: monai still have 3.8.. so lets not move to 3.9

runtests.sh
```diff
@@ -44,39 +44,26 @@ doQuickTests=false
 doNetTests=false
 doDryRun=false
 doUnitTests=false
-doBlackFormat=false
```
SachidanandAlle, line 47: need changes in azure pipeline (which runs everything including tests on GPU)
https://github.com/Project-MONAI/MONAILabel/blob/main/.azure/azure-build-test-gpu-pipeline.yml#L26-L37

SachidanandAlle: Fixed

SachidanandAlle (changes requested):
All looks good to me.. but we have to rollback to 3.8 py version to be in sync with monai (docker, libraries)

However for slicer plugin, if we want to make sure 3.9 needs to be supported, we can add additional check for plugin code (static checks etc..) as additional task/action to be run py 3.9 version

jamesobutler:
I've update the CI to run matrix of Python 3.8 and Python 3.9 for the pythonapp.yml only.

jamesobutler:
Required status check for the build/build-docs/packaging need to be updated now that there are Python 3.8 and 3.9 versions which give the checks different names.

SachidanandAlle (approved):

jamesobutler:
@SachidanandAlle was there a reason to not rebase all these organized commits? This PR was merged with a squash and merge action even though it contained a set of separate changes in various commits. Re comment at https://github.com/Project-MONAI/MONAILabel/pull/736#issuecomment-1094419482.

SachidanandAlle:
Sorry.. i forgot about your specific request for this PR..

PR #740, opened 2022-04-11
Linting updates multiple commits (separate commits)

jamesobutler:
This is a follow-up to #736.

A request for this PR would be to integrate all of the commits in this PR (rebase and merge) to maintain the separate set of changes in their own organized commits rather than squashed down into one large commit which makes it difficult when reviewing the commit history to find out which diff change corresponds to which set of changes and also makes it difficult to cherry-pick certain commits when working with branches in forks.

https://github.com/Project-MONAI/MONAILabel/commit/ba9e89ed1dd7100e9c7b01aad344f187b2673c16 is being reverted in a commit here as a way to avoid rewriting git commit history of protected branch `main`.


SachidanandAlle (approved):

jamesobutler:
Thanks for the rebase and merge!

PR #763, opened 2022-04-29
Adding GMM-based Scribbles method

### LINKED ISSUE #439: Scribbles - Implement a likelihood model using Gaussian Mixture Model
We currently have a histogram-based likelihood model for estimating appearance-based proposals for foreground and background.
While this works for most of the sample apps, it may be more relevant to implement and include more advanced likelihood models (see some discussion here: https://github.com/Project-MONAI/MONAILabel/issues/432#issuecomment-932309545)

For example, the original GrabCut paper (https://www.microsoft.com/en-us/research/wp-content/uploads/2004/08/siggraph04-grabcut.pdf) uses a Gaussian Mixture Model (GMMs) which may be useful and a bit more advanced than histograms

There are a number of implementations for GMMs out there, perhaps the most relevant ones will be the GPU implementation from MONAI and scikit-learn's CPU implementation:
- MONAI (GPU): https://docs.monai.io/en/latest/networks.html?highlight=gmm#monai.networks.layers.GaussianMixtureModel 
- Scikit-learn (CPU): https://scikit-learn.org/stable/modules/generated/sklearn.mixture.GaussianMixture.html

GMM can also naturally extend to multi-class problems, so it will be relevant for when we extend scribbles for multi-class.

masadcv:
Adding Gaussian Mixture Model based likelihood estimator for Scribbles
Fixes #439 

This covers the following:
- Adding GMM based likelihood method from MONAI (GPU)
- Adding GMM based likelihood method from scikit-learn as fallback (CPU)
- Adding Unittests for new feature (GMM likelihood)
- Refactoring Scribbles InferTask to provide `ScribblesLikelihoodInferTask` task to collect common functions
- Adding new dependency (scikit-learn) to requirements.txt
- Moving dependencies (ninja) needed to run MONAI's GMM to requirements.txt from requirements-dev.txt
- Add GMM method to radiology app
- Tested everything with scribbles examples using Spleen Radiology app


monailabel/scribbles/infer.py
```diff
@@ -16,30 +16,21 @@
...
     ApplyGraphCutOptimisationd,
     MakeISegUnaryd,
+    MakeLikelihoodFromScribblesGMMd,
     MakeLikelihoodFromScribblesHistogramd,
 )
 from monailabel.transform.post import BoundingBoxd, Restored
 
 
-class HistogramBasedGraphCut(InferTask):
+class MyScribblesLikelihoodInferTask(InferTask):
```
SachidanandAlle, line 25: can we name it better? Instead of Myxxx

masadcv: Sure, how about `ScribblesLikelihoodInferTask`?

masadcv: Changed it to `ScribblesLikelihoodInferTask`

monailabel/scribbles/utils.py
```diff
@@ -12,6 +12,9 @@
 
 import numpy as np
 import numpymaxflow
+import torch
+from monai.networks.layers import GaussianMixtureModel
+from sklearn.mixture import GaussianMixture
```
tvercaut, line 17: It would be better to avoid an additional dependency (sklearn).

monailabel/scribbles/utils.py
```diff
@@ -160,3 +165,115 @@ def make_likelihood_image_histogram(image, scrib, scribbles_bg_label, scribbles_
...
+        # this may fail if MONAI Cpp Extensions are not loaded properly
+        retprob = learn_and_apply_gmm_monai(
+            image=image,
+            scrib=scrib,
+            scribbles_bg_label=scribbles_bg_label,
+            scribbles_fg_label=scribbles_fg_label,
+            num_mixtures=num_mixtures,
+        )
+    except:
+        # if MONAI's GMM fails, fallback to scikit-learn's GMM
```
tvercaut, line 265: It would be better to have a MONAI-based fallback. If the issue is that the current MONAI GMM is a cuda or c++ extension that may not be available, maybe we should have a second python-only fallback in MONAI. An implementation for it can for example be borrowed from https://github.com/ldeecke/gmm-torch

masadcv: This fallback is because GMM in MONAI only supports CUDA tensors:
https://github.com/Project-MONAI/MONAI/blob/13314320c4c85e2f3da50a71e3d8ec54befe7cd7/monai/_extensions/gmm/gmm_cpu.cpp#L25

Thanks for pointing to gmm-torch library. If I have bandwidth, I will look at this. I also have a OpenCV-GMM that I did back in my uni days :see_no_evil: (https://github.com/masadcv/OpenCVGMM) - not entirely sure if this is accurate or up-to-date. Will check..

monailabel/scribbles/utils.py
masadcv: Agreed! Perhaps for now we can remove this dependency and raise an error when tensors are not on GPU (we still have the other histogram-based method that should be usable on CPU).

monailabel/scribbles/utils.py
tvercaut: Right, I had not looked in the `gmm_cpu.cpp` file. It's indeed more than a packaging issue.

masadcv: There may still be a packaging issue with this on Windows - like we had for SimpleCRF - as JIT would need to compile the source when MONAI GMM is called...

masadcv:
This PR now only utilises MONAI's GMM functions 
The external scikit-learn based functions and dependency have now been removed from this.

This method may only work on GPU (as GMM in MONAI is only currently available for GPU) - I believe this may not be an issue since we already have GPU available for other MONAI Label AI-assisted functions. As a backup we still have Histogram method for users who may only have access to CPU. I believe support for CPU for GMM in MONAI may improve in future (see: https://github.com/Project-MONAI/MONAI/issues/1601#issuecomment-1113668957)

The GMM method is specifically good for cases which are challenging to solve with Histogram. It also does not require a full ROI selection (although such selection is still possible and may get best results). The method can work with few background scribbles (unlike HIstogram method that required an ROI bounding box selection).

masadcv:
Regarding the second dependency `ninja`, I have raised this issue with MONAI here (https://github.com/Project-MONAI/MONAI/issues/4206) and this may be available as an optional install option for MONAI once this PR (https://github.com/Project-MONAI/MONAI/pull/4207) is merged.

SachidanandAlle (approved):

masadcv:
@SachidanandAlle can we merge this pr... many thanks!

Issue #433, opened 2021-09-30
Scribbles - Improve ROI interface 


masadcv:
As suggested in comment here: https://github.com/Project-MONAI/MONAILabel/issues/422#issuecomment-926729860

We may benefit from having a dedicated ROI select button (similar to foreground/background click buttons) that can enable ROI to be placed into the image volume. At the moment, there is quite a bit of navigation involved before a user can access the ROI tool (especially in the latest Slicer preview release).

Additionally this will give us some control over display properties of the ROI element, especially now that in latest Slicer preview these are by default shown in red color, which is easily confused with background scribbles color. Something similar to this:https://github.com/Project-MONAI/MONAILabel/issues/422#issuecomment-927141895 is needed for our case.

masadcv:
Moving this to backlog as I do not have bandwidth to attempt this before v0.3 release.

masadcv:
closing as this was address in #798 by @adamaji

Issue #422, opened 2021-09-24
ROI selection tool broken in Slicer-4.13.0-2021-09-23 preview


masadcv:
We use ROI selection tool for selecting bounding box for scribbles method. It worked great in previous versions of Slicer (e.g. from two days ago in **Slicer-4.13.0-2021-09-21** it worked fine).

With the latest preview release (**Slicer-4.13.0-2021-09-23**), there seems to be two issues:
1. The ROI tool has been significantly changed. It is now under markup module I believe where I am having a difficult time initialising and marking up an `vtkMRMLAnnotationROINode` node (see screenshot of before and after update). It would be good to have some guidance/documentation on how we can do the same interaction that we did in the version from two days ago. If this already exists, can you please point me to it. 
2. The markup module crashes Slicer as soon as I click on ROI symbol from within the tool (see screenshot below)
![Screenshot from 2021-09-24 15-41-21](https://user-images.githubusercontent.com/3410225/134693679-34bdb267-1042-48e4-a8ad-b44dbad427ea.png)
In the terminal from where I run Slicer, it gives the following error messages:

> Switch to module:  "MONAILabel"
  Switch to module:  "Markups"
  GetItemByDataNode: Invalid data node to find
  GetItemByDataNode: Invalid data node to find
  error: [/home/user/Downloads/Slicer-4.13.0-2021-09-23-linux-amd64/bin/SlicerApp-real] exit abnormally - Report the problem.

## Screenshots
Before the update (version **Slicer-4.13.0-2021-09-21** )
![Screenshot from 2021-09-24 15-36-40](https://user-images.githubusercontent.com/3410225/134693003-1ecd93cf-1c82-4dbb-a0f1-2b1fe4bf70ff.png)

After the update (version **Slicer-4.13.0-2021-09-23**)
![Screenshot from 2021-09-24 15-38-58](https://user-images.githubusercontent.com/3410225/134693293-e4d61e34-c134-4252-9b2c-62df5c5c7c2f.png)

cc: @lassoan

lassoan:
Thank you for the feedback and sorry for the disruption. In yesterday's big commit two major changes were made:
- the new markups toolbar was introduced, to clearly separate creation of new markups from placinc control points
- placement of _markups_ ROI nodes is emphasized over legacy _annotation_ ROI nodes

[...]

PR #798, opened 2022-05-20
Add ROI placement widget for scribbles in Slicer plugin

adamaji:
Took a crack at #433 as I'm working on some other ROI tools here. These changes add an ROI placement widget for scribbles in the Slicer plugin so the interface is more direct and there's no reliance on externally defined ROIs.

Right now the GUI looks like this for me on my windows machine as all I've done is add a new row for the markup placement widget. I think probably it would be suited better in the second row so there is more of a progression of operations from top to bottom, but looking for some input here. Perhaps some of the other UI elements could be rearranged as well.
image

Additionally the ROI display is hardcoded to be colored more neutral with an empty fill as described in https://github.com/Project-MONAI/MONAILabel/issues/422#issuecomment-927141895
image

cc @masadcv

### DIFF
--- a/plugins/slicer/MONAILabel/MONAILabel.py
+++ b/plugins/slicer/MONAILabel/MONAILabel.py
@@ -2100,7 +2100,7 @@ class MONAILabelWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
         scribblesROINode.AddObserver(scribblesROINode.PointPositionDefinedEvent, self.onROIPointPlaced)
         scribblesROINode.SetName("Scribbles ROI")
         scribblesROINode.CreateDefaultDisplayNodes()
-        scribblesROINode.GetDisplayNode().SetFillOpacity(0.0)
+        scribblesROINode.GetDisplayNode().SetFillOpacity(0.4)
         scribblesROINode.GetDisplayNode().SetSelectedColor(1, 1, 1)
         scribblesROINode.GetDisplayNode().SetColor(1, 1, 1)
         scribblesROINode.GetDisplayNode().SetActiveColor(1, 1, 1)
--- a/plugins/slicer/MONAILabel/MONAILabel.py
+++ b/plugins/slicer/MONAILabel/MONAILabel.py
@@ -1926,15 +1926,10 @@ class MONAILabelWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
         v.GetRASToIJKMatrix(RasToIjkMatrix)
 
         roi_points_ras = [0.0] * 6
-        if roiNode.__class__.__name__ == "vtkMRMLMarkupsROINode":
-            # for vtkMRMLMarkupsROINode
-            center = [0] * 3
-            roiNode.GetCenter(center)
-            roi_points_ras = [(x - s / 2, x + s / 2) for x, s in zip(center, roiNode.GetSize())]
-            roi_points_ras = [item for sublist in roi_points_ras for item in sublist]
-        else:
-            # if none found then best to return empty list
-            return []
+        center = [0] * 3
+        roiNode.GetCenter(center)
+        roi_points_ras = [(x - s / 2, x + s / 2) for x, s in zip(center, roiNode.GetSize())]
+        roi_points_ras = [item for sublist in roi_points_ras for item in sublist]
 
         min_points_ras = [roi_points_ras[0], roi_points_ras[2], roi_points_ras[4], 1.0]
         max_points_ras = [roi_points_ras[0 + 1], roi_points_ras[2 + 1], roi_points_ras[4 + 1], 1.0]
--- a/plugins/slicer/MONAILabel/MONAILabel.py
+++ b/plugins/slicer/MONAILabel/MONAILabel.py
@@ -1866,7 +1866,9 @@ class MONAILabelWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
 
             # try to get roi if placed
             roiNode = self.ui.scribblesPlaceWidget.currentNode()
-            selected_roi = self.getROIPointsXYZ(roiNode)
+            selected_roi = []
+            if roiNode and roiNode.GetControlPointPlacementComplete():
+                selected_roi = self.getROIPointsXYZ(roiNode)
 
             # send scribbles + label to server along with selected scribbles method
             params = self.getParamsFromConfig("infer", scribblesMethod)
--- a/plugins/slicer/MONAILabel/MONAILabel.py
+++ b/plugins/slicer/MONAILabel/MONAILabel.py
@@ -1931,15 +1931,10 @@ class MONAILabelWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
         roi_points_ras = [0.0] * 6
         if roiNode.__class__.__name__ == "vtkMRMLMarkupsROINode":
             # for vtkMRMLMarkupsROINode
-            print(roiNode.__class__.__name__)
             center = [0] * 3
             roiNode.GetCenter(center)
             roi_points_ras = [(x - s / 2, x + s / 2) for x, s in zip(center, roiNode.GetSize())]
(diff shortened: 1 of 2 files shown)


SachidanandAlle:
This is great..  thanks for the contribution..
Code/logic  looks to me..
@masadcv can you please verify  the plugin at your end..

masadcv:
Many thanks! Will have a closer look at this tomorrow.

One minor point to better utilise widget space, can you move ROI widget to empty space next to Model: Histogram+Graphcut?

169624551-3c79348e-e4b1-425b-8ee6-d3deccbf0f07

adamaji:
I've updated the UI so it looks like this in my environment:

image

plugins/slicer/MONAILabel/MONAILabel.py
```diff
@@ -2084,6 +2093,21 @@ def updateBrushSize(self, value):
             effect = self._scribblesEditorWidget.activeEffect()
             effect.setParameter("BrushAbsoluteDiameter", value)
 
+    def createScribblesROINode(self):
+        scribblesROINode = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsROINode")
+        scribblesROINode.AddObserver(scribblesROINode.PointPositionDefinedEvent, self.onROIPointPlaced)
+        scribblesROINode.SetName("Scribbles ROI")
+        scribblesROINode.CreateDefaultDisplayNodes()
+        scribblesROINode.GetDisplayNode().SetFillOpacity(0.0)
```
masadcv, line 2101: Can you update this to have slight opacity to better differentiate inside and outside of the ROI?

e.g.
`scribblesROINode.GetDisplayNode().SetFillOpacity(0.4)` works well on my end.

plugins/slicer/MONAILabel/MONAILabel.py
```diff
@@ -1981,6 +1985,11 @@ def onResetScribbles(self):
         # remove "scribbles" segments from label
         self.onClearScribblesSegmentNodes()
 
+        # re-add ROI node if removed
+        if self.ui.scribblesPlaceWidget.currentNode() is None:
+            scribblesROINode = self.createScribblesROINode()
```
masadcv, line 1990: This is not the right place for re-initialising - please do it in the `onStartScribbling()` function or create a separate `initialiseROINodes()` function that gets called everytime a new sample is loaded...

plugins/slicer/MONAILabel/MONAILabel.py
```diff
@@ -2084,6 +2093,21 @@ def updateBrushSize(self, value):
             effect = self._scribblesEditorWidget.activeEffect()
             effect.setParameter("BrushAbsoluteDiameter", value)
 
+    def createScribblesROINode(self):
+        scribblesROINode = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsROINode")
+        scribblesROINode.AddObserver(scribblesROINode.PointPositionDefinedEvent, self.onROIPointPlaced)
```
masadcv, line 2098: You are adding observers everytime this function is called, however I do not see a RemoveObserver logic anywhere? Maybe a good idea to cover that along with a `resetROI()` function....

plugins/slicer/MONAILabel/MONAILabel.py
```diff
@@ -1168,9 +1175,10 @@ def onNextSampleButton(self):
                 "Are you sure to continue?"
             ):
                 return
-            self.onResetScribbles()
             slicer.mrmlScene.Clear(0)
 
+        self.onResetScribbles()
```
masadcv, line 1180: I am not sure why you are moving this outside the if condition?
Is there a specific logic that requires this?

masadcv (changes requested):
Many thanks for looking into this. Overall, it looks quite good, however there are a few points to cover in my comments.

At the moment, it works fine if I add ROI with scribbles. However, without ROI, the scribbles seems to be broken. See this where I attempt this: https://www.youtube.com/watch?v=NuwwBTsWDrI  -  can you fix this issue? Possibly you need to return `None` for ROIs when no ROI is placed. This logic is currently missing which may be causing this.  

The second issue is missing `ResetROI()` functionality that needs to be utilized when loading new sample as well as changing labels in multilabel apps. Specifically the ROIs are not reset when new label is selected for multilabel app. See this example where I change label: https://www.youtube.com/watch?v=gGLLn9s_uKo - expected behaviour here is that the ROI be removed when I select kidney as my new label for scribbles.

Give me a ping once these are addressed, I will then give it another try! Let me know if anything is unclear...

plugins/slicer/MONAILabel/Resources/UI/MONAILabel.ui
```diff
@@ -645,6 +645,30 @@
         <item row="2" column="3">
          <widget class="QComboBox" name="scribLabelComboBox"/>
         </item>
+        <item row="0" column="6">
+         <layout class="QHBoxLayout" name="horizontalLayout_4">
+          <item>
+           <widget class="QLabel" name="scribblesROILabel">
+            <property name="text">
+             <string>Place ROI:</string>
```
masadcv, line 653: Lets rename this to `ROI:` 

It may be understood that this is for placing ROI...

adamaji:
Thanks for the review @masadcv . I believe I've addressed all your comments with the latest push here. The scribbles ROI now acts more like the existing segmentation node, in that it only persists across one sample and is kept in a member variable. The issue with scribbles not working without an ROI was because the ROI node was defined, but returned bounds with lengths of 0, so added another check for control point definition there.

One thing I had trouble with was understanding some of the intended function behavior -- for example `clearScribbles` mentions that it is used for resetting tools to default but directly under that function is a `resetScribbles` function, so it's not clear to me how "resetting" scribbles is different from "clearing" here.

masadcv:
> Thanks for the review @masadcv . I believe I've addressed all your comments with the latest push here. The scribbles ROI now acts more like the existing segmentation node, in that it only persists across one sample and is kept in a member variable. The issue with scribbles not working without an ROI was because the ROI node was defined, but returned bounds with lengths of 0, so added another check for control point definition there.

Many thanks!! I will give this a go tomorrow morning and if all looks good, we can proceed with this PR... 

> One thing I had trouble with was understanding some of the intended function behavior -- for example `clearScribbles` mentions that it is used for resetting tools to default but directly under that function is a `resetScribbles` function, so it's not clear to me how "resetting" scribbles is different from "clearing" here.

“Resetting” functionality removes everything associated with scribbles setup (e.g. when you click next you can clear everything and start fresh)

"Clearing" functionality only clears the segmentation labels in scribbles label maps - while retaining all the nodes used by scribbles, this is used in multi-label case where after fine-tuning one label a user switches to a different label but still wants to work with scribbles on the same volume (hence no need to reinitialize these nodes)

I agree the comments may need to be updated and more descriptive in capturing these behaviors...

plugins/slicer/MONAILabel/MONAILabel.py
masadcv: New logic looks good to me..

plugins/slicer/MONAILabel/MONAILabel.py
```diff
@@ -1907,15 +1928,10 @@ def getROIPointsXYZ(self, roiNode):
         roi_points_ras = [0.0] * 6
         if roiNode.__class__.__name__ == "vtkMRMLMarkupsROINode":
```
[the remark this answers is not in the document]
adamaji: Makes sense, pushed an update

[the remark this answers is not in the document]
masadcv: Thanks!

masadcv:
@adamaji 
Awesome! All looks good and works well on my setup.. lets wait for the checks to finish...

Thank you for looking into this and going through the review!

SachidanandAlle:
Adding ROI node is a cleaner solution rather than picking the first node added.
Thank you everyone for your contribution and making it better.

PR #799, opened 2022-05-22
[WIP] Add collaborative (default) and competetive annotation modes

### LINKED ISSUE #787: Feature request: Support multiple annotations on the same image instance
To create high-quality datasets, it is necessary to have the same image annotated multiple times by different people (e.g. double or triple segmentations). At the moment there is no interface for this in MONAILabel.

My team could also contribute to this, but I am not sure what is the best way to implement this feature. Certainly some changes need to be made to the datastore, filter functions and plugins. Also some sort of user management is then necessary.

Currently, for multiple annotations, we use a workaround with multiple instances of MONAILabel sharing the same app. Each instance is then used by a different annotator. Later, the data must then be merged in a cumbersome way.

I would prefer that this functionality be integrated into MONAILabel in a bundled way. I am relatively sure that other teams will need this feature as well (@LilianaCaldeira, @dtruhn, @kbressem).

jlvahldiek:
This PR introduces _collaborative_ (default) and _competetive_ annotation modes to active learning strategies.

- collaborative mode (default)
  - This mode corresponds to the previous behavior. Every saved label contributes to the total workload with no respect to the Client/User-ID.
- competetive mode
  - In this mode every saved label gets tagged by the Client/User-ID as defined in MONAILabel settings in 3D Slicer.
  - The workload progress in 3D Slicer is presented accordingly. 
  - Using this mode it is possible that one image is annotated by several different annotators (they only need a different Client/User-ID). Thus, it is possible to obtain data for an inter-reader reliability analysis.
  - But also data for an intra-reader reliability analysis can be obtained. An individual annotator only needs to use two different client/user IDs for this (e.g. AnnotatorName-1, AnnotatorName-2).

I added a competetive first strategy to the radiology sample app.

This functionality will only work with `LocalDatastore` and 3D Slicer.

I have not changed training implementation. Annotations saved using _competitive_ mode are probably not used for training with the current implementation (not tested).


SachidanandAlle:
The initial logic looks good to me.. so basically we like to have 2 kinds of users/roles.. this logic should scale/continue when we actually add user/roles in monailabel..

jlvahldiek:
Thanks for the review, @SachidanandAlle!

We have been using the proposed competitive mode successfully for a few days now in order to retrieve annotations for an inter-reader analysis.

Of course, this is only a small workaround and we certainly need much more changes (user roles, etc.). But for now this works for the purposes of my group and I wanted to share the code with others.

We don't necessarily have to merge this PR either. On the other hand, this PR has no breaking changes and may add some small enhancements...

PR #821, opened 2022-06-02
Add MONAILabelReviewer Slicer Extension

jlvahldiek:
Introduces an additional Slicer extension called MONAILabelReviewer as discussed in https://github.com/Project-MONAI/MONAILabel/issues/722. This tool allows for quality control, scoring and easy browsing and filtering of annotated datasets and easy search for specific cases. The goal of MONAILabelReviewer is to become a comprehensive review tool that helps in the creation of very high-quality datasets through MONAILabel.

See PR review comments from predecessor PR #771.

This PR will close https://github.com/Project-MONAI/MONAILabel/issues/722.

PR #822, opened 2022-06-02
Monai label reviewer packaging fixes

jamesobutler:
This is a follow-up to my comment at https://github.com/Project-MONAI/MONAILabel/pull/821#issuecomment-1145410374.

A couple commits are to better organize MONAILabelReviewer based on Slicer standards. The important commit here is `BUG: Fix MONAILabelReviewer module not packaged in Slicer extension` so that MONAILabelReviewer is actually packaged correctly as part of the extension. It had not been added to the main CMakeLists.txt and neither had the supporting Lib files for it. This commit makes sure this new module is available when installing the extension from the Slicer extension's manager.

Here is an image of MONAILabelReviewer now in the module finder widget in Slicer. Previously MONAILabelReviewer was not present upon installing the MONAILabel extension.
image

`ENH: Use Slicer standard warning display dialog` is some minor Slicer optimization work so an extra UI file wasn't needing to be included in the package. There is likely other optimization improvements, but this work is a start to at least have the module included in the extension package.

cc: @SachidanandAlle rebasing and merging to include the 4 commits would be helpful to the git history instead of squashing down into a single commit such as for understanding the diff needed to make sure the module is included in the package separately from some refactoring work.


jamesobutler:
@SachidanandAlle What's this failed check? It's unclear how the failure is related here.
image

SachidanandAlle (approved):

PR #842, opened 2022-06-17
Do not truncate node_name

pwrightkcl:
Store node name in full to allow individual elements to truncate it as needed. PR #841 elides text in the 'Master Volume:' dropdown.
Note: on testing, the elided node name appears in slice view, but the dropdown text still appears full length, or at least longer than that in the slice view.

### DIFF
--- a/plugins/slicer/MONAILabel/MONAILabel.py
+++ b/plugins/slicer/MONAILabel/MONAILabel.py
@@ -1203,7 +1203,7 @@ class MONAILabelWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
             image_id = sample["id"]
             image_file = sample.get("path")
             image_name = sample.get("name", image_id)
-            node_name = sample.get("PatientID", sample.get("name", image_id))[-20:]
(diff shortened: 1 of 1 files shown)


plugins/slicer/MONAILabel/MONAILabel.py
```diff
@@ -1203,7 +1203,7 @@ def onNextSampleButton(self):
             image_id = sample["id"]
             image_file = sample.get("path")
             image_name = sample.get("name", image_id)
-            node_name = sample.get("PatientID", sample.get("name", image_id))[-20:]
```
SachidanandAlle, line 1206: have u checked when patience name is too long..

pwrightkcl: Yes. The elided node name appears in slice view, but the dropdown text still appears full length, or at least longer than that in the slice view.
#841 was supposed to elide the dropdown text, but it seems not to have worked. I don't know Qt so I can't comment further.

SachidanandAlle (approved):

PR #843, opened 2022-06-20
Fix overlapping segment when using DeepEdit-DeepGrow

diazandr3s:

### DIFF
--- a/sample-apps/radiology/lib/configs/deepedit.py
+++ b/sample-apps/radiology/lib/configs/deepedit.py
@@ -43,22 +43,22 @@ class DeepEdit(TaskConfig):
         self.tta_samples = None
 
         # Multilabel
-        # self.labels = {
-        #     "spleen": 1,
-        #     "right kidney": 2,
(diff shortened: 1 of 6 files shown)


sample-apps/radiology/lib/configs/deepedit.py
```diff
@@ -43,23 +43,23 @@ def init(self, name: str, model_dir: str, conf: Dict[str, str], planner: Any, **
...
+            "right kidney": 2,
+            "left kidney": 3,
+            "liver": 6,
+            "stomach": 7,
+            "aorta": 8,
+            "inferior vena cava": 9,
             "background": 0,
         }
 
+        # Single label
```
SachidanandAlle, line 57: Are we making multi label as default..?

diazandr3s: Yes, I think it is the generic case. It covers both single and multilabel

SachidanandAlle (approved):

PR #860, opened 2022-07-04
Fix hardcoded label file extension in Slicer extension

jlvahldiek:
PR fixes a hardcoded file extension in `download_label()` of the Slicer extension. Label file extension is now read from Slicer's MONAILabel settings and still defaults to `.nii.gz` in case it is not specified in the settings.


plugins/slicer/MONAILabel/MONAILabelLib/client.py
```diff
@@ -168,7 +169,8 @@ def download_label(self, label_id, tag):
                 MONAILabelError.SERVER_ERROR, f"Status: {status}; Response: {response}", status, response
             )
 
-        local_filename = tempfile.NamedTemporaryFile(dir=self._tmpdir, suffix=".nii.gz").name
+        file_ext = slicer.util.settingsValue("MONAILabel/fileExtension", ".nii.gz")
```
lassoan, line 172: This is OK (an improvement over using a hardcoded extension), but it is not very robust. The setting is shared between multiple Slicer instances that are running at the same time, which may interfere with each other. For example, a user may start some lengthy batch processing operation with one MONAILabel server and in the meantime working on some other data sets interactively on another MONAILabel server, using another Slicer instance. If they use different file formats then one of them may break.

Instead, it would be much more robust and not much more complicated to get the file extension from the file content. It is trivial for NRRD and DICOM:
- NRRD: first 4 bytes of the file is `NRRD`
- DICOM: 4 characters at position 128 are `DICM`

(For NIFTI, as usual, things are more complicated, there is no easy or efficient way to tell that if a file is a NIFTI image. It may be a gzipped file then you need to start with unzipping, then get the "magic string" from the header and then compare it to a list of known valid values. The Python reader that you already use can probably check this, but it may not be efficient - e.g., if it unzipped the entire file for this check then that would be probably an unacceptable overhead).

It would also make sense for the server to provide a filename with a correct extension (for example via `filename` in the `Content-Disposition` header).

SachidanandAlle, line 172: i would rather not bring slicer import here.. let it be pure client api related functions (which later can be kind of common library for all py clients)

instead you can take the argument file_ext for `def download_label(self, label_id, tag)` and it can be passed from slicer related class/lib

another thing is, as @lassoan mentioned, this is for download.. which means the file extension is already correctly sent in `filename` from the server. 

https://github.com/Project-MONAI/MONAILabel/blob/main/monailabel/endpoints/datastore.py#L138


you can use the same for saving.. in that case the logic can be something like..
```
def download_label(self, label_id, tag, file_ext=".nii.gz")
  file_ext = "".join(pathlib.Path(file.filename).suffixes) if file.filename else file_ext
```

SachidanandAlle (changes requested):
as mention in the review comment

jlvahldiek:
I have very limited resources at the moment, unfortunately. I will try to work on the PR in the coming week.

jlvahldiek:
I apologize for the delay. Thanks for your comments. I removed the slicer import and the implementation now gets the correct file extension from the response header.

SachidanandAlle (approved):

PR #865, opened 2022-07-07
Basic XNAT Datastore with list and image fetch

SachidanandAlle:
- list images for given project
- fetch image from xnat

These should be good enough for first integration to support MONAILabel + XNAT for infer related actions/features.


diazandr3s (approved):
Excellent! LGTM

PR #877, opened 2022-07-15
Fix issue with original labels

diazandr3s:


SachidanandAlle (approved):

diazandr3s:
> I thought the fix is needed in the 3d slicer plugin Please check it again

I tested and the issue was that the labels attibute was a list and not a dictionary

PR #879, opened 2022-07-19
Rc4 fixes

SachidanandAlle:
- use array and meta from MetaTensor
- no need of To/From MetaTensor transform
- add integration tests to verify wsi inference over deepedit/nuclick models


wyli (approved):
This update uses the latest metatensor API

PR #884, opened 2022-07-24
Allow custom python

pieper:
The need for this came up in the context of installing monai label in Slicer's python environment where the interpreter executable is called `PythonSlicer`

This change allows users to set an environment variable to select the python version.

### DIFF
--- a/monailabel/scripts/monailabel
+++ b/monailabel/scripts/monailabel
@@ -17,8 +17,8 @@ DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." >/dev/null 2>&1 && pwd)"
 export PYTHONPATH=$DIR:$PYTHONPATH
 echo Using PYTHONPATH=$PYTHONPATH
 
-PYEXE=python
-version=$(python --version 2>&1)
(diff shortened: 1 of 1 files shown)


SachidanandAlle (approved):

monailabel/scripts/monailabel
```diff
@@ -17,8 +17,8 @@ DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." >/dev/null 2>&1 && pwd)"
 export PYTHONPATH=$DIR:$PYTHONPATH
 echo Using PYTHONPATH=$PYTHONPATH
 
-PYEXE=python
-version=$(python --version 2>&1)
+PYEXE=${MONILABEL_PYEXE:-python}
```
lassoan, line 20: There is a typo here `MONILABEL_PYEXE` should be probably `MONAILABEL_PYEXE`.

diazandr3s: Thanks, @lassoan
Fixed this here https://github.com/Project-MONAI/MONAILabel/pull/894

PR #896, opened 2022-07-26
Fix post transform

diazandr3s:


sample-apps/radiology/lib/trainers/segmentation.py
```diff
@@ -101,7 +101,7 @@ def train_pre_transforms(self, context: Context):
     def train_post_transforms(self, context: Context):
         return [
             EnsureTyped(keys="pred", device=context.device),
-            Activationsd(keys="pred", softmax=len(self._labels) > 1, sigmoid=len(self._labels) == 1),
+            Activationsd(keys="pred", softmax=True),
```
diazandr3s, line 104: Even for the single label tasks, we're using softmax as the output network is set in 2

SachidanandAlle (approved):

Issue #908, opened 2022-08-04
Scribbles functionality seems to be broken after MetaTensor update


masadcv:
**Describe the bug**
Running scribbles (Histogram+GraphCut), get the following error:
```
  File "/data/MONAILabel-Dev/monailabel/scribbles/utils.py", line 121, in make_histograms
    values = image[scrib == scribbles_bg_label]
IndexError: boolean index did not match indexed array along dimension 1; dimension is 185 but corresponding boolean dimension is 205
```

**To Reproduce**
1. Follow steps to setup and run radiology app
2. Add scribbles and update

**Expected behavior**
It should work without a problem.

**Additional context**
The scribbles methods expect `image` and `label` (providing scribbles).
When I look through the data transforms for scribbles, some transforms are throwing away MetaTensor information.

For example, `AddBackgroundScribblesFromROId` which adds scribbles from ROI (at: https://github.com/Project-MONAI/MONAILabel/blob/main/monailabel/scribbles/infer.py#L67). Below are affine matrix before and after this transform is applied for the image and label MetaTensor

Before:
```
d["image"].affine
tensor([[   0.9023,    0.0000,    0.0000, -461.0978],
        [   0.0000,    0.9023,    0.0000, -461.0978],
        [   0.0000,    0.0000,    5.0000,    0.0000],
        [   0.0000,    0.0000,    0.0000,    1.0000]], dtype=torch.float64)
d["label"].affine
tensor([[   0.9023,    0.0000,   -0.0000, -461.0978],
        [   0.0000,    0.9023,   -0.0000, -461.0978],
        [   0.0000,    0.0000,    5.0000,    0.0000],
        [   0.0000,    0.0000,    0.0000,    1.0000]], dtype=torch.float64)
```

After:
```
d["image"].affine
tensor([[   2.5000,    0.0000,    0.0000, -461.0978],
        [   0.0000,    2.5000,    0.0000, -461.0978],
        [   0.0000,    0.0000,    5.0000,    0.0000],
        [   0.0000,    0.0000,    0.0000,    1.0000]], dtype=torch.float64)
d["label"].affine
tensor([[2.5000, 0.0000, 0.0000, 0.0000],
        [0.0000, 2.5000, 0.0000, 0.0000],
        [0.0000, 0.0000, 5.0000, 0.0000],
        [0.0000, 0.0000, 0.0000, 1.0000]], dtype=torch.float64)
```

[...]

Issue #844, opened 2022-06-21
Improving scribbles fetch data function


masadcv:
In the current implementation data from dictionary is fetched in scribbles flow using helper function from:
https://github.com/Project-MONAI/MONAILabel/blob/main/monailabel/scribbles/transforms.py#L42

While this has not caused any issues currently, the function is really returning a reference to the data object and hence may result in undesirable effects if the reference object is modified by calling function. 

To address, this needs to copy data to avoid any unintentional logical error.

PR #915, opened 2022-08-04
Fix for MetaTensor write issue in scribbles

masadcv:

- Adds a `_save_data()` to `InteractiveSegmentationTransform` to save data to `MetaTensor` when applicable to address #908 
- Fix `_fetch_data()` to address #844 
- Update scribbles unittests
- Remove deprecated `np.bool` and `np.int` in scribbles code

### DIFF
--- a/monailabel/scribbles/transforms.py
+++ b/monailabel/scribbles/transforms.py
@@ -40,7 +40,15 @@ class InteractiveSegmentationTransform(Transform):
         if key not in data.keys():
             raise ValueError(f"Key {key} not found, present keys {data.keys()}")
 
-        return data[key].array if isinstance(data[key], MetaTensor) else data[key]
+        return data[key].array.copy() if isinstance(data[key], MetaTensor) else data[key].copy()
+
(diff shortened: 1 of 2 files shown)


monailabel/scribbles/transforms.py
```diff
@@ -40,7 +40,15 @@ def _fetch_data(self, data, key):
         if key not in data.keys():
             raise ValueError(f"Key {key} not found, present keys {data.keys()}")
 
-        return data[key].array if isinstance(data[key], MetaTensor) else data[key]
+        return data[key].array.copy() if isinstance(data[key], MetaTensor) else data[key].copy()
```
SachidanandAlle, line 43: r u sure u want to copy?

masadcv: What do you recommend as the best alternative here?

This is to address logical error described here: #844 
If I dont copy then the arrays are passed as a reference. Modifying things in calling function could lead to changes in data[key] location too which could possible lead to possible logical errors...

SachidanandAlle (approved):

Issue #921, opened 2022-08-06
NormalizeLabelsInDatasetd to lost Meta Information 


tangy5:
**Describe the bug**
During handling a discussion problem with a user, I found the transformation **NormalizeLabelsInDatasetd** in deepedit has some issues to be compatible with MetaTensor. It seems the meta information after **NormalizeLabelsInDatasetd** will be lost. 

Version: latest MONAI and MONAILABEL release.

**To Reproduce**
```
from monai.transforms import (
    EnsureChannelFirstd,
    LoadImaged,
    SaveImaged,
    Compose,
    Orientationd
)
from monai.apps.deepedit.transforms import (
    NormalizeLabelsInDatasetd,
)

_labels = {'cancer': 1, 'transfer': 2, 'background': 0}

train_transforms = Compose(
    [
        LoadImaged(keys=("label"), reader="ITKReader"),
        NormalizeLabelsInDatasetd(keys=("label"), label_names=_labels),
        EnsureChannelFirstd(keys=("label")),
        Orientationd(keys=("label"), axcodes="RAS"),
        #========================================================================
        SaveImaged(keys="label", output_dir = './', mode='bilinear', output_postfix='label2')
    ]
)

image_path = [{'label': '/home/yuchengt/yuchengt/2022/MONAI/monailabel_projectYT/MONAILabel-main/test_data/HNC/labels/original/NPC_AI_Detectionl_01.nii.gz'}]
train_transforms(image_path)
```
**Behavior**

`isinstance(img, MetaTensor)` will be True, but `img.meta` will lost meta information. 

**`KeyError: 'filename_or_obj'`**

The tested data can be downloaded via the link that the user provided:

https://drive.google.com/file/d/1VSHNW1m3mLZDch60S_LdHF62kUdiCai_/view?usp=sharing

diazandr3s:
Thanks for creating this issue, @tangy5 

I guess the problem is because this transform changes the label MetaTensor by a numpy array here: https://github.com/Project-MONAI/MONAI/blob/dev/monai/apps/deepedit/transforms.py#L104

tangy5:
> 

Cool, will we have a plan to check this issue or create a fixing PR?
Thank you!

diazandr3s:
Yes, I'll propose a PR for this later today :)

PR #923, opened 2022-08-08
Change default dataset - update NormalizeLabelsInDatasetd

diazandr3s:

- Make `NormalizeLabelsInDatasetd` transform compatible with MetaTensor https://github.com/Project-MONAI/MONAILabel/issues/921
- Make `SmartCacheDataset` the default dataset - This avoids errors using large dataset as CacheDataset

### DIFF
--- a/sample-apps/radiology/lib/infers/segmentation.py
+++ b/sample-apps/radiology/lib/infers/segmentation.py
@@ -16,11 +16,9 @@ from monai.transforms import (
     AsDiscreted,
     EnsureChannelFirstd,
     EnsureTyped,
-    GaussianSmoothd,
     KeepLargestConnectedComponentd,
     LoadImaged,
-    NormalizeIntensityd,
-    ScaleIntensityd,
+    ScaleIntensityRanged,
     Spacingd,
 )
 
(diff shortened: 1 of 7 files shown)


sample-apps/radiology/lib/transforms/transforms.py
```diff
@@ -486,3 +488,38 @@ def __call__(self, data: Mapping[Hashable, np.ndarray]) -> Dict[Hashable, np.nda
...
+            label_names: all label names
+        """
+        super().__init__(keys, allow_missing_keys)
+
+        self.label_names = label_names
+
+    def __call__(self, data: Mapping[Hashable, np.ndarray]) -> Dict[Hashable, np.ndarray]:
+        d: Dict = dict(data)
+        for key in self.key_iterator(d):
+            if key == "label":
```
SachidanandAlle, line 509: can we remove this check.. it's redundant..

SachidanandAlle:
Do we need a new pretrained model based on the different set of pre-transform?  the existing one may not work..
in that case, you can push a new model and use the same.. we can keep existing for backward compatible.. (don't override)

sample-apps/radiology/lib/transforms/transforms.py
diazandr3s: Yes, thanks!

diazandr3s:
> Do we need a new pretrained model based on the different set of pre-transform? the existing one may not work.. in that case, you can push a new model and use the same.. we can keep existing for backward compatible.. (don't override)

Agree. Prefer to keep previous pre transforms

sample-apps/radiology/lib/infers/segmentation.py
```diff
@@ -56,8 +56,8 @@ def __init__(
     def pre_transforms(self, data=None) -> Sequence[Callable]:
         return [
             LoadImaged(keys="image", reader="ITKReader"),
-            EnsureTyped(keys="image", device=data.get("device") if data else None),
```
SachidanandAlle, line 59: this can be first.. to push the basic image into GPU.. and the move channel can happen inside the GPU

diazandr3s: Agree

SachidanandAlle (approved):

PR #926, opened 2022-08-10
Minor fixes - refactoring

SachidanandAlle:

- user can pass val_ds in the Request for training.. it can be a file or json array..
- rename pre to cache and add LoadImageExd in pre
- cvat plugin (flip input/output) based on env config
- print version using `monailabel -v` or `monailabel --version`


diazandr3s (approved):
Looks great to me

PR #929, opened 2022-08-10
Add all labels third stage

diazandr3s:

To Do:

- **Third stage optimization:** Don't load the volume each time to perform inference
- **Third stage:** Add batches during training so all labels in the image are considered at the same time


monailabel/tasks/infer/vertebra_pipeline.py
```diff
@@ -65,13 +70,28 @@ def __call__(self, request):
...
         #################################################
-        result_file_third_stage, result_json_third_stage = self.model_segmentation_vertebra(third_stage_request)
+        # Request for third stage
+        third_stage_request = copy.deepcopy(second_stage_request)
+        all_outs = {}
+        all_keys = []
+        for centroid in result_json_second_stage["centroids"]:
+            third_stage_request["centroids"] = [centroid]
+            # TO DO:
+            # 1/ Remove the AsDiscrete transform in third stage infer so we get pre-activation outputs
```
SachidanandAlle, line 83: you can do that based on the request attributes (say skip_xyz) and in the request copy in pipeline you can enable it to skip

SachidanandAlle (approved):
LGTM.. lets fix the speed issues in next PR

monailabel/tasks/infer/vertebra_pipeline.py
diazandr3s: Do you mean skipping the load image every time?

PR #931, opened 2022-08-13
Document Improvement Plan Task1 and 2: Installation Page and Plugin Guide

tangy5:
Hi @SachidanandAlle @Nic-Ma @diazandr3s ,

This is the second draft PR about document improvement plans.
This is about the installation guide page on MONAI Label, some client users mentioned the installation and setup are a bit complicated and not very straightforward. 

Especially on how to find correct version of viewer and add plugins that compatible to latest release of MONAI Label. 

I thinking an idea on providing users the installation page with 3 steps at first look:

install_steps

The current installation page seems great at installing MONAI Label server. I assume most users are more confused about viewers and how to add MONAI Label plugin to the viewer. 

How about we direct users to specific section of each installation step?
**Using reference link to steps sections.** 

Step1: install MONAI Label server
Step2: choose and install correct version of viewer software.
Step3: add plugins to viewer. 

Then guide users to more sample use cases. 

I pushed a very draft version of the installation page, probably we need more art design on this page. But the idea is here.

Thank you! Let me know if we like this idea, we can do more improvement discussion and setup.

Yucheng

The page looks like this:

Screen Shot 2022-08-13 at 3 18 25 PM


SachidanandAlle:
Why there are some 75 commits.. I guess you are missing something on github usage..

PR #940, opened 2022-08-18
Remove install-time openslide dependency

pieper:
In https://github.com/Project-MONAI/MONAILabel/pull/687 openslide was made optional at runtime but it was still in the install dependencies and on some platforms (e.g. ubuntu 20.04) the regular packages don't work for building (pip install fails).  This PR allows the build to finish without openslide.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,7 +10,7 @@
 # limitations under the License.
 
 torch>=1.7
-monai[nibabel, skimage, pillow, tensorboard, gdown, ignite, torchvision, itk, tqdm, lmdb, psutil, openslide, fire]>=0.9.1
+monai[nibabel, skimage, pillow, tensorboard, gdown, ignite, torchvision, itk, tqdm, lmdb, psutil, fire]>=0.9.1
 uvicorn==0.17.6
 pydantic==1.9.1
 python-dotenv==0.20.0
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,7 +36,7 @@ setup_requires =
     ninja
 install_requires =
     torch>=1.7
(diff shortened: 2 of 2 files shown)


SachidanandAlle:
We need this for pathology use-case.. even static compile will fail without this..

help to understand what stopped working.. it should do simple pip install.  rather than going into build mode..  anything changed in latest version of openslide pip package?

pieper:
Even simple `pip install monailabel` fails with openslide dependency.   I've never found a way to make the install work unless this is removed.  Installing the openslide tools with apt-get doesn't make a difference.  Can we make it optional somehow so that it's only required for the pathology apps?  The radiology apps work with this dependency removed.

SachidanandAlle:
Can you provide more details.. day in day out it all works on windows and ubuntu.. you can check the ci builds.. they all passing everytime..

pieper:
I'll close this and assume the openslide install issues are specific to my system.  I thought we should remove it as a hard requirement since the pathology code already has error handling for the case where it's not available.

SachidanandAlle:
you mentioned `pip install monailabel` fails.. but how about pip install openslide-python.  it should also fail in your case..
`pip install openslide-python==1.1.2`

at the end MONAI tries to install this package.. 
https://github.com/Project-MONAI/MONAI/blob/dev/setup.cfg#L83-L84

recently openslide has released some version, but when I tried to install that, it didn't try to build any local native libraries..
checked for windows/ubuntu

trying to make sure we are not missing some actual issue

pieper:
I started from a fresh install and could not reproduce this so I'm not sure if something changed or I missed a step before.

PR #948, opened 2022-08-25
Rename/Move files and delete non required

SachidanandAlle:
- Move files related to deepedit from endoscopy to monailabel core
- Move pipelines to individual app.. deepgrow and vertebra pipeline goes to radiology app
-


diazandr3s (approved):
Looks great to me, thanks!

Issue #945, opened 2022-08-23
DICOMWebDatastore  limited to modality "CT"


jlvahldiek:
The modality that DICOMWebDatastore is searching for is hardcoded and limited to "CT":
https://github.com/Project-MONAI/MONAILabel/blob/48441c3353cd7940a1b9e7e0d5fc06b1b2a9a4c9/monailabel/datastore/dicom.py#L43

It would be very useful for my team to use DICOMWebDatastore with other modalities (MR, CR, etc.). 

What do you think is the best way to make the modality customizable by the user? Maybe it would be also helpful in the future to define some kind of filter list (i.e. modality, date, sex, etc.) that DICOMWebDatastore will use to list images - but this is probably an issue in itself.

SachidanandAlle:
Yes.. this was done long back.. but some improvements in xnat based.. we don't need conversations from dicom to nifti.. dicom data store can be improved..

PR #949, opened 2022-08-25
Improve DICOMWebDatastore

jlvahldiek:
This PR suggests some minor improvements to the DICOMWebDatastore.

- replaces hardcoded search filter "Modality: CT" by a customizable dictionary. This allows to use i.e. other modalities ("Modality: MR") or i.e. filter for specific DICOM series ("SeriesNumber": 42), etc...
- search filter now can be set via a newly introduced setting parameter: MONAI_LABEL_DICOMWEB_SEARCH_FILTER
  - search filter still defaults to "Modality: CT"
- improves listing of labeled images by adding one more check
- changes caching of listing images/labeled images which priorly caused strange behavior

_addresses #945  and #944_


monailabel/datastore/dicom.py
```diff
@@ -109,14 +115,13 @@ def _dicom_info(self, series_id):
             info[f] = str(meta[f].value) if meta.get(f) else "UNK"
         return info
 
-    @cached(cache=TTLCache(maxsize=16, ttl=600))
+    @cached(cache=TTLCache(maxsize=16, ttl=60))
```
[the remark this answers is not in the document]
jlvahldiek: added config parameter MONAI_LABEL_DICOMWEB_CACHE_EXPIRY, defaults to 180 s

monailabel/datastore/dicom.py
```diff
@@ -109,14 +116,13 @@ def _dicom_info(self, series_id):
...
-    @cached(cache=TTLCache(maxsize=16, ttl=600))
+    @cached(cache=TTLCache(maxsize=16, ttl=settings.MONAI_LABEL_DICOMWEB_CACHE_EXPIRY))
     def list_images(self) -> List[str]:
-        datasets = self._client.search_for_series(search_filters={"Modality": self._modality})
+        datasets = self._client.search_for_series(search_filters=self._search_filter)
         series = [str(Dataset.from_json(ds)["SeriesInstanceUID"].value) for ds in datasets]
         logger.debug("Total Series: {}\n{}".format(len(series), "\n".join(series)))
         return series
 
-    @cached(cache=TTLCache(maxsize=16, ttl=600))
```
SachidanandAlle, line 119: Any specific reason why caching is removed.. for next sample.. we do list images - labeled images.. 
this is more helpful for aws (static dicom web storage) where hardly new studies get uploaded

jlvahldiek: when new labels are stored the displayed number of labeled images is not updated frequently enough. sometimes labeling takes only seconds, but the progress is then not visible to the user which is a bit confusing. I agree, for listing just plain images 10 mins is okay (only infrequent changes) - but for listing labeled images I am not sure if caching makes sense (especially for fast labelling tasks). 

Should I make it configurable with a default value of 600?

SachidanandAlle: 3 mins is good.. I agree 10 mins too long :)

SachidanandAlle (approved):

PR #955, opened 2022-08-30
Configurable read timeout of DICOMWeb requests

jlvahldiek:
In a scenario with several hundreds of DICOM series in a DICOMWebDatastore it is possible that OHIF fails to retrieve the whole study list. This is caused by a default timeout of 5 seconds of the used httpx.AsyncClient.

This PR introduces the setting "MONAI_LABEL_DICOMWEB_READ_TIMEOUT" to make the read timeout of default 5.0 seconds extendable via settings parameter.


monailabel/config.py
```diff
@@ -49,6 +49,7 @@ class Settings(BaseSettings):
     MONAI_LABEL_DICOMWEB_FETCH_BY_FRAME: bool = False
     MONAI_LABEL_DICOMWEB_SEARCH_FILTER: Dict[str, Any] = {"Modality": "CT"}
     MONAI_LABEL_DICOMWEB_CACHE_EXPIRY: int = 180
+    MONAI_LABEL_DICOMWEB_READ_TIMEOUT = 5.0
```
SachidanandAlle, line 52: `MONAI_LABEL_DICOMWEB_READ_TIMEOUT: int = 5`

jlvahldiek: Maybe float better? See: https://github.com/encode/httpx/blob/965b8adec3633e1bd249d17fb7a7eed414c74748/httpx/_config.py#L208
```suggestion
    MONAI_LABEL_DICOMWEB_READ_TIMEOUT: float = 5.0
```

SachidanandAlle (approved):

PR #980, opened 2022-09-08
[WIP] Variance Metric Integration From MONAICore | TestCase

finalelement:
Scoping out testcase for Variance Metric that has been integrated in MONAILabel, but it needs to be tested.

We should follow the philosophy "If it's not tested, it's broken"


finalelement:
Will re-open this later.

PR #978, opened 2022-09-07
Variance Metric From MONAI Core Integration to MONAI Label for all Apps

finalelement:
Integrating to the particular 'monailabel/tasks/scoring/epistemic_v2.py'

If there are other places where modifications to be made for scoring @SachidanandAlle please suggest.

Signed-off-by: Vishwesh Nath <vnath@nvidia.com>

PR #991, opened 2022-09-13
Fix network for tooltracking based on monai rc2 + optimize transform

SachidanandAlle:

Input Image size: (1920, 1080, 3)
Input Model: Tool Tracking

- Total e2e Latency = 180 ms
- Pre Transform Latency = 60 ms (30 ms for Load; 20 ms for resize)
- Infer Latency = 40 ms
- Post Transform = 80 ms (80 ms for Find Contour)

Exclude Find Contour in Scoring Task (epistemic)... 15% speed up in overall time


SachidanandAlle:
@finalelement the previous PR is not correctly verified: https://github.com/Project-MONAI/MONAILabel/pull/978

`VarianceMetric` from monai core...
It doesn't work and throws error for NAN values.. so did kind of rollback to use np.nanvar (based on function arg)

Check for options in torch.var to ignore NaN values

diazandr3s (approved):
LGTM

PR #993, opened 2022-09-13
update bundle app readme

tangy5:

Update Readme file with a table to instruct users with proper models from model-zoo. Highlighted labeling models and bundle. (Since not all models in the model-zoo are appropriate to use in MONAI Label) @Nic-Ma, let me know if this makes sense, we will not recommend some non-labeling bundle to use in the Bundle app in MONAI Label, such as MedNIST_GAN.

### DIFF
--- a/sample-apps/monaibundle/README.md
+++ b/sample-apps/monaibundle/README.md
@@ -33,6 +33,25 @@ However the following constraints has to be met for any monai bundle to directly
 ### Overview
 
 
+### Supported Models 
+
+The Bundle App supports most labeling models in the Model Zoo, please see the table for labeling tasks.
+
+
+| Bundle | Model | Objects | Modality | Note |
+|:----:|:-----:|:-------:|:--------:|:----:|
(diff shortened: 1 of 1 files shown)


sample-apps/monaibundle/README.md
```diff
@@ -50,9 +69,6 @@ monailabel start_server --app workspace/monaibundle --studies workspace/images -
 # Pick DeepEdit And Segmentation model (multiple models)
 monailabel start_server --app workspace/monaibundle --studies workspace/images --conf models "spleen_ct_segmentation_v0.1.0,spleen_deepedit_annotation_v0.1.0"
```
SachidanandAlle, line 70: is it still spleen_ct_segmentation_v0.1.0 or spleen_ct_segmentation

SachidanandAlle (approved):

sample-apps/monaibundle/README.md
tangy5: It's should be with version "_v0.1.0" or "_v0.1.1", users need to specify which bundle version to load.

Nic-Ma: Hi @tangy5 ,

Maybe you can leverage the MONAI bundle `download` API with `bundle name` and `verson`:
https://github.com/Project-MONAI/MONAI/blob/dev/monai/bundle/scripts.py#L175

Thanks.

PR #1027, opened 2022-09-29
Support random strategy for next patch for WSI

SachidanandAlle:
Signed-off-by: SACHIDANAND ALLE <sachidanand.alle@gmail.com>

This is to support basic random strategy for Pathology WSI images

- you can select random WSI image
- you can select random path from the existing WSI image
- QuPath UI Changes to fetch Next Random Patch


diazandr3s (approved):
LGTM

Issue #1014, opened 2022-09-24
Bug in function download_label


manigalati:
Hello! :)
I ran into the following bug while working with the plugin of monailabel in slicer.

**Describe the bug**
The application always fails at loading original labels, with the following message appearing in the log:
"Original label not found ... "

**To Reproduce**
After running batch inference of any model, original labels are created inside the server. When trying to fetch the next sample from the server in slicer, original labels never appear.

**Solution**
I debugged the code myself. The code always raises an exception that enters inside the catch block at line 1245 of file 
https://github.com/Project-MONAI/MONAILabel/blob/main/plugins/slicer/MONAILabel/MONAILabel.py 

The reason is to be found at line 173 of https://github.com/Project-MONAI/MONAILabel/blob/main/plugins/slicer/MONAILabel/MONAILabelLib/client.py

Here, the method getheader is invoked on the response object only after being read at line 357 of the same file. But response.read() turns the response from being an http response object to be a simple bytes object encapturing the content of the response. On a bytes object, getheader cannot be called, and this is what triggers an exception.

I hope I have been clear enough, so that my message can be useful to solve the issue quickly.

Best regards,
Francesco

diazandr3s:
Hi @manigalati 

Thanks for raising this issue.

> The application always fails at loading original labels, with the following message appearing in the log:
> "Original label not found ... "

By default, MONAI Label searches in the labels/original folder for the predicted label for the current volume (https://github.com/Project-MONAI/MONAILabel/blob/main/plugins/slicer/MONAILabel/MONAILabel.py#L1234-L1245). If it doesn't exist, it will show that warning.

See the following image the flag `Original Labels` is activated. 

![Screenshot from 2022-09-27 00-14-15](https://user-images.githubusercontent.com/11991079/192397379-b02ac88a-48e5-47e3-a468-c1ed505dd30d.png)

You can get to this in Slicer on Edit -> Application Settings

> After running batch inference of any model, original labels are created inside the server. When trying to fetch the next sample from the server in slicer, original labels never appear.

How did you perform batch inference? Did you use the REST API on the web browser? Have you checked that the predictions are being generated on labels/original folder?

Hope that helps

SachidanandAlle:
@tangy5 can you please check.. may be it needs something to be fixed on 3D slicer side..

tangy5:
> @tangy5 can you please check.. may be it needs something to be fixed on 3D slicer side..

Good, I will look into this.

[...]

PR #1031, opened 2022-09-30
Fix downlad_label() to load labels from Original directory

tangy5:

The PR to address issue: https://github.com/Project-MONAI/MONAILabel/issues/1014

Error when load labels from original directory. "When trying to fetch the next sample from the server in slicer, original labels never appear." 

The issue is caused when calling response.getheadet(content_disposition) after response.read().

Moved the call of response.getheadet(content_disposition) in send_response and return the content_disposition , the download_label function works normal.

Let me know if the solution is good, there might be other solutions. Welcome discuss.

### DIFF
--- a/plugins/slicer/MONAILabel/MONAILabelLib/client.py
+++ b/plugins/slicer/MONAILabel/MONAILabelLib/client.py
@@ -164,13 +164,12 @@ class MONAILabelClient:
         selector = "/datastore/label?label={}&tag={}".format(
             MONAILabelUtils.urllib_quote_plus(label_id), MONAILabelUtils.urllib_quote_plus(tag)
         )
-        status, response, _ = MONAILabelUtils.http_method("GET", self._server_url, selector)
+        status, response, content_disposition = MONAILabelUtils.http_method("GET", self._server_url, selector)
         if status != 200:
             raise MONAILabelException(
                 MONAILabelError.SERVER_ERROR, f"Status: {status}; Response: {response}", status, response
             )
 
-        content_disposition = response.getheader("content-disposition")
         file_name = re.findall('filename="(.+)"', content_disposition)[0]
         file_ext = "".join(Path(file_name).suffixes)
         local_filename = tempfile.NamedTemporaryFile(dir=self._tmpdir, suffix=file_ext).name
@@ -342,6 +341,7 @@ class MONAILabelUtils:
         logging.debug(f"HTTP Response Headers: {response.getheaders()}")
 
         response_content_type = response.getheader("content-type", content_type)
+        content_disposition = response.getheader("content-disposition")
         logging.debug(f"HTTP Response Content-Type: {response_content_type}")
 
         if "multipart" in response_content_type:
@@ -351,10 +351,10 @@ class MONAILabelUtils:
                 logging.debug(f"Response FILES: {files.keys()}")
                 return response.status, form, files
             else:
-                return response.status, response.read(), None
+                return response.status, response.read(), content_disposition
 
         logging.debug("Reading status/content from simple response!")
-        return response.status, response.read(), None
+        return response.status, response.read(), content_disposition
 
     @staticmethod
     def save_result(files, tmpdir):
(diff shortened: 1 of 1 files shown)


diazandr3s (approved):
LGTM

plugins/slicer/MONAILabel/MONAILabelLib/client.py
```diff
@@ -351,10 +351,10 @@ def send_response(conn, content_type="application/json"):
                 logging.debug(f"Response FILES: {files.keys()}")
                 return response.status, form, files
             else:
-                return response.status, response.read(), None
+                return response.status, response.read(), content_disposition
 
         logging.debug("Reading status/content from simple response!")
-        return response.status, response.read(), None
+        return response.status, response.read(), content_disposition
```
SachidanandAlle, line 357: The 3rd tuple in the current returned value is files.. so you can add 4th tuple to send all the response headers (instead of just content disposition).. and fetch the content disposition from the list of response headers


```python
        if "multipart" in response_content_type:
            if response.status == 200:
                form, files = MONAILabelUtils.parse_multipart(response.fp if response.fp else response, response.msg)
                logging.debug(f"Response FORM: {form}")
                logging.debug(f"Response FILES: {files.keys()}")
                return response.status, form, files, response.getheaders()
            else:
                return response.status, response.read(), None, response.getheaders()

        logging.debug("Reading status/content from simple response!")
        return response.status, response.read(), None, response.getheaders()

```

SachidanandAlle (changes requested):
add new return tuple in for `send_response` and use the same across

plugins/slicer/MONAILabel/MONAILabelLib/client.py
```diff
@@ -342,19 +341,20 @@ def send_response(conn, content_type="application/json"):
         logging.debug(f"HTTP Response Headers: {response.getheaders()}")
 
         response_content_type = response.getheader("content-type", content_type)
+        content_disposition = response.getheader("content-disposition")
```
SachidanandAlle, line 344: Return response headers.. that way users can use what they want

plugins/slicer/MONAILabel/MONAILabelLib/client.py
tangy5: This is correct, thanks, the multipart function used the 3rd tuple for returning files. I've recover the 3rd and added 4th for sending response headers.

tangy5: > This is correct, thanks, the multipart function used the 3rd tuple for returning files. I've recover the 3rd and added 4th for sending response headers.  All returning variables using send_response are modified. 
Let me know if this is a decent solution.

plugins/slicer/MONAILabel/MONAILabelLib/client.py
tangy5: Yup, response.headers is returned and users can get key-values from the dict.

plugins/slicer/MONAILabel/MONAILabelLib/client.py
```diff
@@ -164,14 +164,13 @@ def download_label(self, label_id, tag):
...
-        status, response, _ = MONAILabelUtils.http_method("GET", self._server_url, selector)
+        status, response, _, headers = MONAILabelUtils.http_method("GET", self._server_url, selector)
         if status != 200:
             raise MONAILabelException(
                 MONAILabelError.SERVER_ERROR, f"Status: {status}; Response: {response}", status, response
             )
 
-        content_disposition = response.getheader("content-disposition")
-        file_name = re.findall('filename="(.+)"', content_disposition)[0]
+        file_name = re.findall('filename="(.+)"', headers["content-disposition"])[0]
```
SachidanandAlle, line 173: this is list of tuples.. it's something like [(header_name, header_value]+

tangy5: Verified.

SachidanandAlle (approved):

PR #1019, opened 2022-09-27
Update app and proxy for GCP integration

diazandr3s:
Signed-off-by: Andres <diazandr3s@gmail.com>

PR #1039, opened 2022-10-04
Revert "Update app and proxy for GCP integration"

SachidanandAlle:
Reverts Project-MONAI/MONAILabel#1019

```python
@router.get("/dicom/{path:path}", include_in_schema=False)
async def proxy(path: str, response: Response, user: User = Depends(get_basic_user)):
    return await proxy_dicom("", path, response)```

this breaks the existing connecting/functionality with other dicom web (e.g. orthanc) over ohif


SachidanandAlle:
@tangy5 and @diazandr3s next time, please make sure basic things are verified
meanwhile in future, lets try to add more integration functions to catch such errors..

PR #1040, opened 2022-10-04
Config to Enable/Disable DICOM => NIFTI conversion

SachidanandAlle:

When using OHIF only, then we don't need to convert DICOM to NIFTI
In such cases user can bypass DICOM to NIFTI conversion as follows:

```bash
export MONAI_LABEL_DICOMWEB_CONVERT_TO_NIFTI=false
monailabel start_server --app apps/radiology --studies http://127.0.0.1:8042/dicom-web
```


diazandr3s (approved):
LGTM

PR #1041, opened 2022-10-04
Add method to download the entire labeled dataset as a ZIP archive

jlvahldiek:
### Motivation:
Especially when working with the DICOMweb datastore, there is currently no convenient functionality to download the entire dataset (including image and label files).

### Aim:
Add functionality to download a ZIP archive containing all images and corresponding label files in research format via API request.

### Changes:
- add datastore API endpoint `api_download_dataset()`
- add method `download_dataset()` to generate and return a ZIP archive that contains image and corresponding label files
- the number of included cases can be limited by `limit_cases` parameter


monailabel/endpoints/datastore.py
```diff
@@ -181,6 +182,28 @@ def update_label_info(label: str, tag: str, info: str = Form("{}"), user: Option
...
+        for d in dl:
+            # write image and corresponding label file to archive
+            for key in d.keys():
+                path = d[key]
+                archive.write(path, arcname=os.path.join(key, os.path.basename(path)))
+    assert archive.filename is not None
+    return FileResponse(archive.filename, filename=os.path.basename(archive.filename))
+
+
 @router.get("/", summary="Get All Images/Labels from datastore")
```
SachidanandAlle, line 207: Option 1::
GET /datatstore?include_images=False
(We can use the existing API with additional URI parameter to define this instead of introducing new API)

Option 2::
GET /datatstore/dataset as new API to download the images+labels+dataset.json for research purposes 

Current name is little confusing 
GET /datatstore/downloadlabeleddataset

I feel option 2 is better.. later variations can be added to same endpoint to support different formats etc..

monailabel/endpoints/datastore.py
```diff
@@ -181,6 +182,28 @@ def update_label_info(label: str, tag: str, info: str = Form("{}"), user: Option
...
+        logger.info(f"Number of cases in datalist reduced to: {limit_cases} of {len(dl)} case(s)")
+        dl = dl[:limit_cases]
+
+    with zipfile.ZipFile("dataset.zip", mode="w") as archive:
+        logger.info(f"ZIP archive will be written to: {archive.filename}")
+        for d in dl:
+            # write image and corresponding label file to archive
+            for key in d.keys():
+                path = d[key]
+                archive.write(path, arcname=os.path.join(key, os.path.basename(path)))
```
SachidanandAlle, line 202: Better we define an abstract method in dataset interface and let derived classes provide the implementation.. some support file.. some support http.. and some can throw NotSupported exception..

jlvahldiek:
Thank you for your thoughts, @SachidanandAlle. Will work on option 2.

SachidanandAlle (approved):
Thank you :)

PR #1043, opened 2022-10-04
Support load local bundle to monaibundle app

tangy5:

Add support loading local bundle if the model is not in Model-Zoo.

The logic is:
1 First check --conf models <bundle>, the bundle is in Model-Zoo, if available, use model-zoo model.
2. If not, check whether the bundle is in local model directory, if no, exit.
3. Then monaibundle app load models either from Model-Zoo or local.

BTW, let's think about whether we still need "all" as an option, because not all Model-Zoo bundle are suitable to use in MONAI Label, e.g., non labeling bundle such as MedNISTGAN. We might not want to suggest users to load "all" bundles into MONAI Label. 

Thanks

### DIFF
--- a/sample-apps/monaibundle/main.py
+++ b/sample-apps/monaibundle/main.py
@@ -58,23 +58,34 @@ class MyApp(MONAILabelApp):
 
         models = models.split(",")
         models = [m.strip() for m in models]
-        invalid = [m for m in models if m != "all" and not available.get(m)]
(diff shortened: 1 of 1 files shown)


SachidanandAlle (approved):

SachidanandAlle:
We can remove all option.. please raise a separate PR for the same.. (Possibly u need to update the documents as well after removing the all option)

PR #1057, opened 2022-10-10
Fix parsing of variables/settings

jlvahldiek:
In #1048 some setting variables were defined to default to `NoneType`.

During `start_server()` these variables are parsed and re-set from `NoneType` to `str` 
 because of `str(None) = "None"`. This leads to malforming of the resulting Qido-, Wado- and Stow-Prefixes and wrong URLs.

This PR fixes the parsing of `NoneType` settings.

Co-authored by @markus-hinsche

### DIFF
--- a/monailabel/main.py
+++ b/monailabel/main.py
@@ -325,9 +325,14 @@ class Main:
             logger.debug("                  ENV VARIABLES/SETTINGS                  ")
             logger.debug("**********************************************************")
             for k, v in settings.dict().items():
-                v = json.dumps(v) if isinstance(v, list) or isinstance(v, dict) else (str(v) if v else None)
+                if isinstance(v, list) or isinstance(v, dict):
+                    v = json.dumps(v)
+                elif v is not None:
+                    v = str(v)
+                else:
+                    v = None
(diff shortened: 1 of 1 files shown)


monailabel/main.py
```diff
@@ -325,9 +325,15 @@ def start_server_init_settings(self, args):
             logger.debug("                  ENV VARIABLES/SETTINGS                  ")
             logger.debug("**********************************************************")
             for k, v in settings.dict().items():
-                v = json.dumps(v) if isinstance(v, list) or isinstance(v, dict) else str(v)
+                if isinstance(v, list) or isinstance(v, dict):
+                    v = json.dumps(v)
+                elif v is not None:
+                    v = str(v)
+                else:
+                    v = None
```
jlvahldiek, line 333:
```suggestion
```
Could be ommitted.

markus-hinsche: I agree, feel free to delete the else block

SachidanandAlle (commented):
In some cases we use array.. we need json conversation.. you can hard check for "None" and not convert for such value

SachidanandAlle (approved):

PR #1060, opened 2022-10-11
Fix Interface + Add deprecation

SachidanandAlle:

- Fix Interface for InferTask (deprecate and use BasicInferTask instead)
- Make get_annotations_by_image_id specific for DSADatastore (others doesn't have any idea)
- Deprecate Endpoints to Run All Train/Scoring (encourage to use model/method specific task for both training/scoring)
- Remove support for ALL in 3D Slicer under TrainTasks
- Remove --conf models all (for bundle app) and force user to use specific model name(s) from model zoo


diazandr3s (approved):
LGTM

PR #1063, opened 2022-10-11
Remove references to Orthanc from Dockerfile

jlvahldiek:
References in MONAILabel's Dockerfile imply that the container ships with Orthanc included. This may be confusing.
_See also: #594_


SachidanandAlle (approved):

diazandr3s:
> References in MONAILabel's Dockerfile imply that the container ships with Orthanc included. This may be confusing. _See also: #594_

Thanks for this, @jlvahldiek!

PR #1098, opened 2022-10-26
Updated formatted string syntax so it will work with Python v3.7

qing-at-work:
Minor update to formatted string syntax so it will work with Python v3.7


SachidanandAlle (approved):

SachidanandAlle:
Thank you for the contributions to make it better

PR #1106, opened 2022-10-30
Fix building of OHIF in Docker image

jlvahldiek:
OHIF never gets built in Docker image due to a misplaced build argument. This PR re-activates the default building of OHIF in MONAILabel's Docker image.

See also https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact and https://stackoverflow.com/a/56748289

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,10 +14,10 @@
 # to use different version of MONAI pass `--build-arg MONAI_IMAGE=...`
 
 ARG MONAI_IMAGE=projectmonai/monai:1.0.1
-ARG BUILD_OHIF=true
 
 FROM ${MONAI_IMAGE} as build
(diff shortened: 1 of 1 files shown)


SachidanandAlle (approved):

SachidanandAlle:
@tangy5 we can have some OHIF related tests in integration (which we run after we build the docker).. this is to make sure OHIF prebuilt is released as part of it (even file/path check for those compiled files related to OHIF)

tangy5:
> @tangy5 we can have some OHIF related tests in integration (which we run after we build the docker).. this is to make sure OHIF prebuilt is released as part of it (even file/path check for those compiled files related to OHIF)

Sure, this makes a lot of sense, made the OHIF related tests on my working plate.

PR #1110, opened 2022-11-02
Add endoscopic bundle name option

tangy5:
Small change for endoscopic models config: tooltracking and inbody.
Add conf options so that users can load/initiate tooltracking/inbody models with any name in local workspace. 
This is partly for MONAI Toolkit loading bundle from NGC, they can have other names if not download from Github.

### DIFF
--- a/sample-apps/endoscopy/lib/configs/inbody.py
+++ b/sample-apps/endoscopy/lib/configs/inbody.py
@@ -34,7 +34,7 @@ class InBody(TaskConfig):
         super().init(name, model_dir, conf, planner, **kwargs)
 
         bundle_name = conf.get("bundle_name", "endoscopic_inbody_classification")
-        bundle_version = "0.2.0"
+        bundle_version = conf.get("bundle_version", "0.2.0")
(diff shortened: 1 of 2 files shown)


sample-apps/endoscopy/lib/configs/inbody.py
```diff
@@ -33,7 +33,7 @@ class InBody(TaskConfig):
     def init(self, name: str, model_dir: str, conf: Dict[str, str], planner: Any, **kwargs):
         super().init(name, model_dir, conf, planner, **kwargs)
 
-        bundle_name = "endoscopic_inbody_classification"
+        bundle_name = conf.get("bundle_name", "endoscopic_inbody_classification")
         bundle_version = "0.2.0"
```
SachidanandAlle, line 37: in that case, you also take bundle_version (similar to bundle_name).. name alone is not enough in this case..

tangy5: Sure, thanks!

SachidanandAlle (approved):

PR #1172, opened 2022-12-03
Using Pathology bundles for nuclick and classification models

SachidanandAlle:


tangy5 (approved):
Looks good. 
We might want to use bundle for pathology in Toolkit as well in next release.

PR #1183, opened 2022-12-06
Add Detection Task Support - Infer

tangy5:
This PR adds detection task support for MONAI Label.

Inference pipeline supported. 

Feature:

MONAI Label + 3D Slicer for detection annotation.

How to start: 

- User can try the inference with MSD Task06 Lung nodule dataset:

`monailabel datasets --download --name Task06_Lung --output .
`
- Then start monailabel server: 
`monailabel start_server --app sample-apps/monaibundle/ --studies Task06_Lung/imagesTr --conf models lung_nodule_ct_detection_v0.5.0`

- Use the new MONAI Label Slicer plugin and developer mode to load plugin. 

Inference result and display as Box ROI in Slicer: 

detection_demo

Video demo for a usecase: https://drive.google.com/file/d/1UXIht9x7vUM9P_JGIAhBp_Bg6_BfCG7j/view?usp=share_link

Some note:

- Latest detection bundle: `**lung_nodule_ct_detection_v0.5.0**` is used for development.
- A "detector" function is used, as the detector is a wrapper function of inferer, so treated the detector as a separate InferType. 
- run_detector function should accommodate detection task data format, and wrap network, RetinaInferer, box selector in detector.
- DetectionWriter is for generating detection result json.
- create_slicer_detection_json is used for adapting 3d Slicer MarkupsROINode.
- Tested latest repo with integration test, tested several latest maintained bundles.

### DIFF
--- a/monailabel/tasks/infer/basic_infer.py
+++ b/monailabel/tasks/infer/basic_infer.py
@@ -558,9 +558,10 @@ class BasicInferTask(InferTask):
             return cw(data)
 
         if self.type == InferType.DETECTION:
-            writer = DetectionWriter(pred_box_key=self.target_box_key, pred_label_key=self.target_label_key)
-        else:
-            writer = Writer(label=self.output_label_key, json=self.output_json_key)        
+            dw = DetectionWriter(pred_box_key=self.target_box_key, pred_label_key=self.target_label_key)
+            return dw(data)
+        
+        writer = Writer(label=self.output_label_key, json=self.output_json_key)        
             
         return writer(data)
 
--- a/monailabel/tasks/infer/basic_infer.py
+++ b/monailabel/tasks/infer/basic_infer.py
@@ -20,12 +20,13 @@ from typing import Any, Callable, Dict, List, Sequence, Tuple, Union
 import torch
 from monai.data import decollate_batch
 from monai.inferers import Inferer, SimpleInferer, SlidingWindowInferer
+from monai.apps.detection.networks.retinanet_detector import RetinaNetDetector
 
 from monailabel.interfaces.exception import MONAILabelError, MONAILabelException
 from monailabel.interfaces.tasks.infer_v2 import InferTask, InferType
 from monailabel.interfaces.utils.transform import dump_data, run_transforms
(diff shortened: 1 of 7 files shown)


monailabel/tasks/infer/basic_infer.py
```diff
@@ -297,7 +302,11 @@ def __call__(
         latency_pre = time.time() - start
 
         start = time.time()
-        data = self.run_inferer(data, device=device)
+        data = (
+            self.run_inferer(data, device=device)
+            if not self.type == InferType.DETECTION
+            else self.run_detector(data, device=device)
```
SachidanandAlle, line 308: what it makes different to run infer vs detector.  at the end both should probably run network in eval mode.. is not true?

i thought detector is just another model which outputs some predictions and as part of post processing, we do generate ROI/BoundingBox as the output?  am i missing something here?

tangy5: That's the tricky thing here. 

The detection model defined in the bundle has this:

https://github.com/Project-MONAI/model-zoo/blob/1be315fe132f5071830c07617e91d619868559f5/models/lung_nodule_ct_detection/configs/inference.json#L61

The detector is a wrapper of higher concept of inferer in detection task, it contains the networks, box selectors, anchor_generator, inferer, etc. 

The detector can have different networks. 

That's what I understand from the detection bundle side. 
Maybe @Can-Zhao can explain more of this, she helped me understand the design of "detector". :)

tangy5: To help us understand the detection task:

Here is the difference of Detector and Detection Network in MONAI Core:

Detector: 
https://github.com/Project-MONAI/MONAI/blob/6fa4bce1961d84f89c243f27c093054396ee4557/monai/apps/detection/networks/retinanet_detector.py#L65

Detection Network:
https://github.com/Project-MONAI/MONAI/blob/6fa4bce1961d84f89c243f27c093054396ee4557/monai/apps/detection/networks/retinanet_network.py#L199

SachidanandAlle (approved):

Issue #1198, opened 2022-12-09
Bad Monailabel annotations in DSA on EC2 Amazon Linux


orenlivne:
**Describe the bug**
The annotations produced by the `segmentation_nuclei` model in the monailabel DSA plugin are nonsensical, on EC2 Amazon Linux.

**To Reproduce**
1. Create an EC2 GPU instance with Amazon Linux OS: `Linux ip-172-31-65-118.ec2.internal 5.10.147-133.644.amzn2.x86_64 #1 SMP Fri Oct 14 01:16:24 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux`.
3. Install DSA using docker-compose.
4. Build openslide from sources.
5. Install monailabel from github sources.
6. Install monailabel DSA plugin using the Slicer CLI Web plugin on the admin dashboard.
7. Open a test image.
8. Submit a monai annotation job.

**Screenshots**
<img width="1265" alt="Screen Shot 2022-12-09 at 12 42 01 PM" src="https://user-images.githubusercontent.com/2911829/206761790-ae06b1f9-cd37-4235-92bc-f49c5a7923a8.png">

Note that the same workflow works on MacOS. Annotations are good there.

SachidanandAlle:
Thanks for reporting.. we will look into the issue.. and provide the fix asap..

@tangy5

tangy5:
Pathology application recently has big changes, I guess we haven't test DSA much. Let me look into this. 
Thanks for reporting the issue.

SachidanandAlle:
We will work on this and provide the solution asap (next couple of days).  It will be included in release candidate planned by end of the week.

Meanwhile you can try to use release version 0.5.2 if you are blocked on something.

PR #1200, opened 2022-12-11
Fix DSA

tangy5:
The DSA use case is impacted by our recent changes on `FindContoursd` transform in the post-processing for QuPath application.  Linked to the reported issue (https://github.com/Project-MONAI/MONAILabel/issues/1198).

`p = np.moveaxis(p, 0, 1)  # for cv2
`

in `FindContoursd`

We can keep the changes of `FindContoursd`, to gurantee the QuPath use case. Then make modification at `DSA writer` to flip contours. 

Changes been made in the PR. 

**The branch is tested on both DSA and QuPath, both are working correctly now.** 

DSA now works:

Screenshot from 2022-12-11 12-01-12

### DIFF
--- a/monailabel/utils/others/pathology.py
+++ b/monailabel/utils/others/pathology.py
@@ -12,6 +12,7 @@
 import json
 import logging
 import tempfile
+
 import numpy as np
 
 from monailabel.utils.others.label_colors import to_hex, to_rgb
@@ -40,7 +41,7 @@ def create_dsa_annotations_json(json_data, loglevel="INFO"):
 
             color_map = annotation.get("labels")
             elements = annotation.get("elements", [])
-            location = annotation.get("location", []) 
+            location = annotation.get("location", [])
 
             for element in elements:
                 label = element["label"]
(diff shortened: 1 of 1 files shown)


SachidanandAlle:
Have you considered changing in DSA plugin instead? If we r doing explicit flip then we fix that instead

SachidanandAlle (approved):
If tested for both DSA and Latest QuPath.. then is all ok .

SachidanandAlle:
I think vanilla PNG fails if you test in QuPath

tangy5:
> Have you considered changing in DSA plugin instead? If we r doing explicit flip then we fix that instead

Thanks, let me see if we can flip at plugin side. If that is better (like flip in CVAT), we can change to that way.

tangy5:
> I think vanilla PNG fails if you test in QuPath

PNG inference failes in QuPath? I tested with WSI. Let me test a PNG

SachidanandAlle:
If it fails... Can you revert the changes and fix the DSA plugin instead

tangy5:
The PNG issue might be another bug? I saw the issue is when selecting a ROI, the entire figure is inferenced instead of the ROI. 

QuPATH inference uses "asap" as outuput type, DSA uses "dsa". The changes in the PR is only with DSA writer.

PR #1208, opened 2022-12-14
use exec to avoid spawning of another process

SachidanandAlle:


diazandr3s (approved):
LGTM

PR #1210, opened 2022-12-15
More UT | Integration Test | Detection Doc | Fix 

tangy5:
Some general thing for 0.6.0 release preparation. 

The  branch is tested with GPU-based unit tests, integration test, passed all tests.

1. Created more unit test for deepedit transform, handler, interaction | More transform unit tests | more endpoint unit tests | Other unit tests.
2. Detection integration test, small fix on training with mlflow.
3. Detection doc.
4. Some updates.

Current GPU Unit Test Coverage reaches a healthier 70%. (from 50%)

UT_coverage

Need to work more on the datastore UT, client UT, corner UT test cases, and some UT for separated functions.


monailabel/tasks/train/bundle.py
```diff
@@ -125,6 +125,9 @@ def _fetch_datalist(self, request, datastore: Datastore):
 
     def _partition_datalist(self, datalist, request, shuffle=False):
         if "detection" in request.get("model"):
+            settings.MONAI_LABEL_TRACKING_ENABLED = (
```
SachidanandAlle, line 128: you don't have to override the settings.. instead you override the `request["tracking"] = "None"` or even empty or None will be fine too..

tangy5: Thanks, made changes according to the prior PR on disabling mlflow for bundles. All bundles works well.

SachidanandAlle (approved):

PR #1217, opened 2022-12-19
NormalizedLabels fix

danieltudosiu:
Minor fix that makes NormalizedLabels agnostic to the order of the labels.


diazandr3s (approved):
LGTM

PR #1251, opened 2023-01-22
Updated segmentation

diazandr3s:


SachidanandAlle:
I will test it first.. please don't merge

SachidanandAlle (approved):
Tested.. and verified across multiple samples.. all looks good and we can replace it with older segmentation model

PR #1350, opened 2023-03-22
Support monai 1.2.0rc2 version

SachidanandAlle:
- Upgrade monai dependency for pip and docker
- Fixes related to write_nifti api removed from latest monai
- Fixes related to deepgrow transform (spatial_dim) argument change from latest monai
- Remove deprecated APIs from schema for openapi


tangy5 (approved):
Looks good.

PR #1365, opened 2023-03-27
Invert orientation in Restored transform

diazandr3s:

### DIFF
--- a/monailabel/transform/post.py
+++ b/monailabel/transform/post.py
@@ -13,12 +13,20 @@ import logging
 from typing import Dict, Hashable, Mapping, Optional, Sequence, Union
 
 import cv2
+import nibabel as nib
 import numpy as np
 import skimage.measure as measure
 import torch
 from monai.config import KeysCollection, NdarrayOrTensor
 from monai.data import MetaTensor
-from monai.transforms import MapTransform, Resize, Transform, generate_spatial_bounding_box, get_extreme_points
+from monai.transforms import (
+    MapTransform,
+    Orientation,
+    Resize,
+    Transform,
+    generate_spatial_bounding_box,
+    get_extreme_points,
+)
 from monai.utils import InterpolateMode, convert_to_numpy, ensure_tuple_rep
 from shapely.geometry import Point, Polygon
 from torchvision.utils import make_grid, save_image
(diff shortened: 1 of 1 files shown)


monailabel/transform/post.py
```diff
@@ -124,6 +134,16 @@ def __call__(self, data):
                 resizer = Resize(spatial_size=spatial_size, mode=self.mode[idx])
                 result = resizer(result, mode=self.mode[idx], align_corners=self.align_corners[idx])
 
+            if self.invert_orient:
+                # Undo Orientation
+                orig_affine = meta_dict["original_affine"]
+                orig_axcodes = nib.orientations.aff2axcodes(orig_affine)
+                inverse_transform = Orientation(axcodes=orig_axcodes)
+                # Apply inverse
+                logging.info(f"THIS IS THE SHAPEEE: {len(result.shape)}")
```
tangy5, line 143: Is the "SHAPEEE" preferred here? Rocky logs :)

diazandr3s: hahahaha great catch! Just updated it

tangy5:
Hi @diazandr3s , thanks for adding the inverted orientation. Do you know if we have relevant test cases that can cover the "invert orientation" case in this transformation? We can think about the unit test in the future for keeping the coverage >80%. Thanks.

diazandr3s:
> Hi @diazandr3s , thanks for adding the inverted orientation. Do you know if we have relevant test cases that can cover the "invert orientation" case in this transformation? We can think about the unit test in the future for keeping the coverage >80%. Thanks.

Good point! Do you think it is better if we move all these to MONAI Core? https://github.com/Project-MONAI/MONAILabel/blob/main/monailabel/transform/post.py#L31

tangy5 (approved):

monailabel/transform/post.py
```diff
@@ -124,6 +134,15 @@ def __call__(self, data):
                 resizer = Resize(spatial_size=spatial_size, mode=self.mode[idx])
                 result = resizer(result, mode=self.mode[idx], align_corners=self.align_corners[idx])
 
+            if self.invert_orient:
+                # Undo Orientation
+                orig_affine = meta_dict["original_affine"]
```
SachidanandAlle, line 139: what if "original_affine" not exists? in that case you want to skip undo-orientation

diazandr3s: Good point. Just updated this. Hope this makes sense :)

PR #1374, opened 2023-04-01
Updated Main Repo README, Sample App READMEs, and Plugin READMEs

zephyrie:
I've updated the main repo README and moved information throughout the repo, so things are hopefully more consistent.

For sample applications, I've updated the main README and each app's README.

For the viewers, I've added a main README and updated all viewers except for DSA.

PR #1376, opened 2023-04-04
Fix ToC links, other links, added broken images, and cleaned up ToC/N…

zephyrie:
- Fixed some issues with the Table of Content links not working
- Fixed a few other links I noticed that weren't working
- Fixed broken images that weren't updated in initial pull
- Cleaned up ToC and moved section to from Endo->CVAT

@tangy5, this should fix the issues from #1374 you mentioned.


tangy5 (approved):
Looks great!

SachidanandAlle (approved):

PR #1380, opened 2023-04-06
Dep version upgrade

SachidanandAlle:
- upgrade dependencies
- Fix if check for numpy array

### DIFF
--- a/monailabel/transform/post.py
+++ b/monailabel/transform/post.py
@@ -137,7 +137,7 @@ class Restored(MapTransform):
             if self.invert_orient:
                 # Undo Orientation
                 orig_affine = meta_dict.get("original_affine", None)
-                if orig_affine:
+                if orig_affine is not None:
                     orig_axcodes = nib.orientations.aff2axcodes(orig_affine)
                     inverse_transform = Orientation(axcodes=orig_axcodes)
                     # Apply inverse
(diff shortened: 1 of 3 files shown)


tangy5 (approved):

YanxuanLiu:
Hi @SachidanandAlle, 
Should we ignore the vulnerability? It's better to fix it if the component is a direct dependence.

SachidanandAlle:
They are again from the plugin folder.. I have mentioned earlier.. we need to discard plugin folder from the scan..

pxLi:
> They are again from the plugin folder.. I have mentioned earlier.. we need to discard plugin folder from the scan..

Again, we are not allowed to do so. 
For now we can only fix vuls of direct dependencies, or add exceptions in internal allow list if indirect dependencies.

SachidanandAlle:
I have no solution for this.. we are looking for exceptions here.. or we have to remove the plugins from this repository

SachidanandAlle:
They are picking these dependencies from QuPath and OHIF base project (external) files.. and we don't control them.. they are other 3rd party oos projects

pxLi:
> I have no solution for this.. we are looking for exceptions here.. or we have to remove the plugins from this repository

We have the options to whitelist those, 
the point here is that please help determine if it could be fixed or its actually a indirect dep, 
so we could take actions but not just ignore those, thanks~

pxLi:
I have added this one to whitelist as its a thirdparty dep and we cannot fix from our-side to unblock CI

PR #1404, opened 2023-04-12
Toolkit NGC bundle compatibilities

tangy5:
As bundle API "get_all_bundles_list" and "get_bundle_version" won't support in NGC. 
We separate the get bundle info logics here so that under NGC env, it will not have github model zoo access.


tangy5:
@yiheng-wang-nv @mingxin-zheng 

Here we will separate the access to github model zoo when only the env var is set to "github".
Underder "ngc", it will not have any github model zoo access. 

Then we can test the download logic in Toolkit.

SachidanandAlle (approved):

PR #1412, opened 2023-04-20
Update LargestCCD to handle things errors better

Geeks-Sid:
Signed-off-by: Siddhesh Thakur <sid.cre8er@gmail.com>

### DIFF
--- a/monailabel/transform/post.py
+++ b/monailabel/transform/post.py
@@ -15,6 +15,7 @@ from typing import Dict, Hashable, Mapping, Optional, Sequence, Union
 import cv2
 import nibabel as nib
 import numpy as np
+import skimage.measure as measure
 import torch
 from monai.config import KeysCollection, NdarrayOrTensor
 from monai.data import MetaTensor
@@ -26,7 +27,6 @@ from monai.transforms import (
     generate_spatial_bounding_box,
     get_extreme_points,
 )
-import skimage.measure as measure
 from monai.utils import InterpolateMode, convert_to_numpy, ensure_tuple_rep
 from shapely.geometry import Point, Polygon
 from torchvision.utils import make_grid, save_image


monailabel/transform/post.py
```diff
@@ -37,28 +37,57 @@
...
+        """
+        Transforms the input data dictionary by applying the get_largest_cc function to the label images.
+
+        Args:
+            data: a dictionary of input data.
+
+        Returns:
+            A dictionary of transformed data with the same keys as the input data.
+        """
+        data = dict()
```
SachidanandAlle, line 86: you are removing all other keys from existing data..
that's the whole reason to have d = dict(data) which shallow copies or type-cast the object

monailabel/transform/post.py
```diff
@@ -37,28 +37,57 @@
 
 
 # TODO:: Move to MONAI ??
-
-
 class LargestCCd(MapTransform):
```
SachidanandAlle, line 40: for which app/version u r using this class.. ?

This will be deprecated soon..
please use `KeepLargestConnectedComponentd` from monai.transforms instead
https://docs.monai.io/en/latest/transforms.html#keeplargestconnectedcomponentd

SachidanandAlle:
Also in the description, you can help to understand what errors you got.. for which use-case you got.. please define your problem with some logs/screenshots if possible

monailabel/transform/post.py
```diff
@@ -37,23 +37,52 @@
...
+            has_channel: whether the data has channel dimension or not.
+        """
         super().__init__(keys)
         self.has_channel = has_channel
 
     @staticmethod
     def get_largest_cc(label):
-        largest_cc = np.zeros(shape=label.shape, dtype=label.dtype)
-        for i, item in enumerate(label):
-            item = measure.label(item, connectivity=1)
```
SachidanandAlle, line 51: i believe the only change u need is

```python
item = ndimage.label(item, connectivity=1)
if item.max() != 0:
  largest_cc[i, ...] = item == ndimage.mode(item, axis=None)[0][0]
```

which is using `ndimage.label` and `ndimage.mode`

if see largest_cc already has the shape of original labels and all initialized to zeros..
so you only pick one label and try to update the corresponding index if the label is valid (max != 0)

monailabel/transform/post.py
```diff
@@ -37,23 +37,52 @@
...
+        for item in label:
+            try:
+                item_label = ndimage.label(item, connectivity=1)
+                if item_label.max() != 0:
+                    item_lcc = item_label == ndimage.mode(item_label, axis=None)[0][0]
+                else:
+                    item_lcc = np.zeros_like(item)
+            except ValueError:
+                item_lcc = np.zeros_like(item)
+            largest_cc.append(item_lcc)
```
SachidanandAlle, line 73: largest_cc = np.zeros_like(label) this will already have the initial shape.. and now you are trying to append..

Can you confirm if you have verified the result... (testing it via 3D Slicer for Deepgrow Pipeline for which this class is currently used)

PR #1466, opened 2023-06-09
0.7.0 release prepare

tangy5:


SachidanandAlle:
Can y cross check previous release. Do we say == or => for monai version?

SachidanandAlle:
Also build/test (epistemic I guess) failing due to lazy param or there is some really a problem at monailabel side

SachidanandAlle (approved):

PR #1476, opened 2023-06-21
Fix md5 hashing with FIPS mode

MattTheCuber:
MD5 hashing is not allowed in FIPS enabled machines for security. A simple fix is to use the `usedforsecurity=False` flag in the `md5()` calls.

### DIFF
--- a/monailabel/datastore/dicom.py
+++ b/monailabel/datastore/dicom.py
@@ -9,12 +9,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import hashlib
 import logging
 import os
 import pathlib
 import shutil
-import sys
 from typing import Any, Dict, Iterator, List, Optional, Tuple
 
 import requests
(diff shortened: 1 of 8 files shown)


SachidanandAlle (commented):
You can simply add a wrapper function in the utility class and use the same.. that will avoid writing same if-else condition too many times.. code/changes will look simpler

MattTheCuber:
> You can simply add a wrapper function in the utility class and use the same.. that will avoid writing same if-else condition too many times.. code/changes will look simpler

Do you mean in https://github.com/Project-MONAI/MONAILabel/blob/main/monailabel/utils/others/class_utils.py?

SachidanandAlle:
https://github.com/Project-MONAI/MONAILabel/blob/main/monailabel/utils/others/generic.py

```python
def md5_digest(s:str) -> str:
    if sys.version_info.minor < 9:
        return hashlib.md5(s.encode("utf-8")).hexdigest()
    return hashlib.md5(s.encode("utf-8"), usedforsecurity=False).hexdigest()
```

MattTheCuber:
Done. _thanks git..._

SachidanandAlle (approved):

PR #1483, opened 2023-07-05
use methods from GeneralTools for getting filename, fix #1480

### LINKED ISSUE #1480: the first `/` is missing if we use `Utils.getFileName` on Linux system
The URI in qupath looks like `file:/home/username/myimage.png` on Linux, and `file:/C:/Users/username/Desktop/myimage.png` on Windows.So `Utils.getFileName` works for Windows but not Linux. We would get `home/username/myimage.png`. This could leads to bugs when we use the result return fromt this method, eg: https://github.com/Project-MONAI/MONAILabel/blob/d95c8a2ea6c8b9dd0bc6d9a06cd3bd8767df3da0/plugins/qupath/src/main/java/qupath/lib/extension/monailabel/commands/RunInference.java#L207-L208
quick fix, insert these lines:
```java
if (System.getProperty("os.name").toLowerCase().contains("linux")) {
	file = "/" + file;
}
```
just before https://github.com/Project-MONAI/MONAILabel/blob/d95c8a2ea6c8b9dd0bc6d9a06cd3bd8767df3da0/plugins/qupath/src/main/java/qupath/lib/extension/monailabel/Utils.java#L34

Maybe we should also check for MacOS or other non-Windows systems? How about this:
```java
if (!System.getProperty("os.name").toLowerCase().contains("windows")) {
	file = "/" + file;
}
```

hubutui:
1. delete `Utils.getFileName` and `Utils.getNameWithoutExtension`.
2. get URI by `imageData.getServer().getURIs()` and then get the filename.
3. get filename without extension by `GeneralTools.getNameWithoutExtension(File)`.

Note: `GeneralTools.getNameWithoutExtension(String)` is not working, see also https://github.com/qupath/qupath/pull/1275


SachidanandAlle:
@hubutui can you please resolve merge conflicts.. also please verify/test one more time at your side for both normal use-case and specific windows use-case 

we should be able to merge this after you do so... 
thanks :)

hubutui:
@SachidanandAlle Hi, I resolved the conflicts and updated the commit. I also tested it in Windows and Linux. It worked as expected.

SachidanandAlle (approved):

Issue #1501, opened 2023-07-25
MonaiLabel Segmentation Error


brchck:
I am trying to train a new dataset (8 bit with 14 channels) within slicer via the MonaiLabel extension. I have adjusted the segmentation config to contain the correct parameters. 

All the images and labels are loading fine. Not sure what other information is needed - any questions please leave a comment and I'll reply ASAP

This is the metadata from the images
`NRRD0005`
`This NRRD file was generated by pynrrd`
`on 2023-07-04 00:07:44(GMT).`
`Complete NRRD file format specification at:`
`http://teem.sourceforge.net/nrrd/format.html`
`type: uint8`
`dimension: 3`
`sizes: 225 285 400`
`encoding: gzip`

Here is the error:
[weird_err_msg_monai.txt](https://github.com/Project-MONAI/MONAILabel/files/12159369/weird_err_msg_monai.txt)

diazandr3s:
Hi @brchck,

Thanks for opening this issue.
Which format are the volumes (i.e. NRRD. NIfTI)?
It seems at least one volume has header issues or you are using thw wrong reader. From the logs I can see this:

```
  File "/g/data/if89/apps/monailabel/0.7.0/lib/python3.9/site-packages/monai/data/image_reader.py", line 1339, in _get_affine
    direction = img.header["space directions"]
KeyError: 'space directions'
The above exception was the direct cause of the following exception:
```

this is the reader method getting the affine from the headers: https://github.com/Project-MONAI/MONAI/blob/2800a76764bc2f58402718fcef8b38f3a60776d0/monai/data/image_reader.py#L1330-L1347

If images are in .nii.gz or .nii format, consider using the itkreader instead (https://github.com/Project-MONAI/MONAI/blob/2800a76764bc2f58402718fcef8b38f3a60776d0/monai/data/image_reader.py#L151). Otherwise I'd suggest you check the volume headers. There might be an issue there.

Hope this helps,

dorkylever:
Hi all, 

I made a pull request: #1505. This is a just a function that I made in util/others that would init the origin and direction in images without that information.

We use a modified version of MPI/LAMA (a mouse embryo registration pipeline) and part of the pre-processing is to remove spacing, orientation and directional information.

It might be a bit far fetched but do think it is possible to do a quick check of the first img's header, then call the function if it's missing info. 

Kind Regards, 
Kyle

PR #1505, opened 2023-07-27
function to initialise orientation and direction in nrrd images

dorkylever:
Just the function - function will eventually need to be called


SachidanandAlle (commented):
I didn't understand why this is needed.. 
where is it used?

dorkylever:
Hi it's in relation to error #1501. 

It's currently not used. I haven't implemented it into the pipeline as I'm not sure about whether to use it in the pipeline (e.g. as a try-except / as a plugin etc.) or keep it as a preprocessing script (i.e., make it a cmd line argument). 

If you don't want to add we can just keep it as a seperate script in our codespace. Just thought this may be a simple error trap that could help others.

monailabel/utils/others/init_direction_and_orient.py
```diff
@@ -0,0 +1,72 @@
...
+    volpaths = get_file_paths(target_dir)
+
+    for i, path in enumerate(volpaths):
+        vol = sitk.ReadImage(path)
+        # pa = sitk.PermuteAxesImageFilter()
+        # pa.SetOrder([1,0,2])
+        # flipped_vol = pa.Execute(vol)
+
+        # sets Direction and Origin
+        vol.SetDirection([-1.0, 0.0, 0.0, 0.0, -1.0, 0.0, 0.0, 0.0, 1.0])
```
lassoan, line 70: Note that spacing is included in 'space directions` in the NRRD file header but in SimpleITK images there is a separate `SetSpacing` method for it.

dorkylever: Right, because we should have the spacing in our output. 

If needed, we can just add vol.SetSpacing([1.0, 1.0, 1.0]) as line 71 and that should do the trick? 
Let me know your opinions.

lassoan: SimpleITK writer will always write the image geometry (origin, spacing, axis directions) to the output image file, I just wanted to mention this to make sure that no information is lost when this code is used in a different context.

The main issue though is that the image geometry should not be set arbitrarily. For example, how do we know that the IJK image axes are RAS (encoded by the -1 values in the direction matrix) and not LPS (direction matrix is identity matrix)? Instead of setting some hardcoded default geometry, it should be copied from the input image.

dorkylever:
Ultimately should be just left outside and be part of a preprocessing script. 
We manually correct our image geometry if needed but other end-users may not.

Issue #1508, opened 2023-07-28
Loading network weights with wrong state_dict fails silently


matt3o:
**Describe the bug**
I spent a few hours wondering why my network won't work and which of my transforms are broken, only to find out that the loading of the network did not work. Monailabel via default uses the "model" key for the state_dict, whereas my code uses the "net" key.
However running MONAILabel, the network is just initialized randomly and the loading of the weights fails completely silently.

The easiest solution would be to change the default value of `load_strict` to `True` in `BasicInferTask`.
This would make a lot of sense for beginners. MONAILabel hides away the specifics of loading the network weights - the least I would expect is an error message telling me something may have gone wrong. In this case however it is certain that something went wrong so it would be even better for the program to fail imo.

tangy5:
Thanks for the suggestions. 
This could be an enhancing feature for the future. The key for loading model weights are not always "state_dict". This can be optimized. 
The reason why the default is not set to "load_strict" to true is that most of the times users may not have all the weights matched. Falise will enable loading parts of pretrained weights. For this feature, we may need a argument that can adjust the configuration. 

Thanks again for the suggestions, if you get time, you could raise a PR for your suggestions, I think they are useful to all users. Or we will put these items in TODO list and enhance in the normal dev plan. Thank you.

matt3o:
Thanks for the quick response @tangy5. Hm yeah I get that for many models you want that value to be False. Why not set it to true in the BasicInferTask and then to False in every sample-app/model, where it is actually needed?
For me as a noobie trying to add a new model this was a major pain.

I can add a PR however I'd need to know which of the two options you prefer:
1) After loading the checkpoint, go through the keys and make sure the desired one is in the state_dict of checkpoint. If not print an error message but continue the program. This would only modify `BasicInferTask._get_network()`. No change in the previous behaviour apart from an error message stating that the weights might not have been loaded correctly or at least were only partial.
2) Set the default value of `load_strict` to `True` in `BasicInferTask` and change the value in all available sample apps where it is not set so far to `False`.

tangy5:
I think the two items you proposed makes sense, feel free to raise a PR. You could add "-s" to your commit command, and our team can edit and test your branch, we will have CI/CD integration and unit tests and env build. I will also try and help if anything is missing or need discussion.

Thanks again.

matt3o:
I'll close this issue now since the code for it was already merged.

[...]

PR #1521, opened 2023-08-08
Fix silent fail if network loads wrong weights

matt3o:
This is a draft for fixing https://github.com/Project-MONAI/MONAILabel/issues/1508 .  It sets load_strict to False in basic_infer.py, also adds a warning in case load_strict is False but the model does not contain all the necessary keys.
Most importanly this PR helps the creation of new models where one might not even realize that the loaded network does not contain the necessary keys, which is what happened to me.

For default the DeepEdit the warning looks like this:
image

It is a bit hidden, but nevertheless if anyone where to debug the code, this should probably help.

Todo:
- [ ] Is this solution a valid solution for everyone?
- [ ] Add load_strict=False to all existing models (only done for DeepEdit so far)

### DIFF
--- a/monailabel/tasks/infer/basic_infer.py
+++ b/monailabel/tasks/infer/basic_infer.py
@@ -452,14 +452,17 @@ class BasicInferTask(InferTask):
 
                 if path:
                     checkpoint = torch.load(path, map_location=torch.device(device))
-                    for key in self.model_state_dict:
-                        if key not in checkpoint:
-                            logger.warning(
-                                f"Expected key {key} has not been found in the checkpoint keys: {checkpoint.keys()}"
-                            )
-                            logger.warning("The run will now continue unless load_strict is set to True")
-                            break
                     model_state_dict = checkpoint.get(self.model_state_dict, checkpoint)
+
+                    if set(self.network.state_dict().keys()) != set(checkpoint.keys()):
+                        logger.warning(
+                            f"Checkpoint keys don't match network.state_dict()! Items that exist in only one dict"
+                            f" but not in the other: {set(self.network.state_dict().keys()) ^ set(checkpoint.keys())}"
+                        )
+                        logger.warning(
+                                "The run will now continue unless load_strict is set to True. "
+                                "If loading fails or the network behaves abnormally, please check the loaded weights"
+                        )
                     network.load_state_dict(model_state_dict, strict=self.load_strict)
             else:
                 network = torch.jit.load(path, map_location=torch.device(device))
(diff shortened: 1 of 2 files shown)


matt3o:
@tangy5 Can you shortly check if this PR is fine so far? Then I would add `self.load_strict = False` to all the other models as well

tangy5:
It looks good to me, Thanks. @SachidanandAlle @diazandr3s , do we have further concerns if we set load_strict to true here by default? Im thinking it should be OK, but I'm not sure about all of our pretrained model weights. Thanks

monailabel/tasks/infer/basic_infer.py
```diff
@@ -57,7 +57,7 @@ def __init__(
         output_label_key: str = "pred",
         output_json_key: str = "result",
         config: Union[None, Dict[str, Any]] = None,
-        load_strict: bool = False,
+        load_strict: bool = True,
```
SachidanandAlle, line 60: lets keep the base class value same.. lets not change.. we override it to true in all derived classes

SachidanandAlle, line 60:
```suggestion
        load_strict: bool = False,
```

matt3o: Then only the warning?
My idea behind that was that this makes adding new models harder. Which is why I proposed changing the default to True and setting the value to False in all the existing models. 
In my case the network would not load any weights at all but MONAILabel would not tell me. The state dict was completely different but I did not even get a warning. I was thus wondering for a few hours why a perfectly fine network wouldn't work in MONAILabel...

SachidanandAlle: In that case can you verify all sample apps..

matt3o: @SachidanandAlle Can you explain what you mean?

SachidanandAlle: we need to make sure existing pre-trained models can load with this flag. 
radiology/pathology and endoscopy apps/models

SachidanandAlle:
I will merge this for now.. but better this gets verified for other existing example models/apps

monailabel/tasks/infer/basic_infer.py
matt3o: @SachidanandAlle Absolutely right. What I had planned was to add that flag in the step, when you guys approve of the changes, which I was not sure of.
However you already merged the branch now. Shall I still add the `flag=False` for all existing models or do you want to do that based on feedback what breaks?

Thanks for the quick response!

matt3o: I will add that today and update this PR / open another one if I can't do that. It might break some preexisting code otherwise.

PR #1524, opened 2023-08-11
Add request info to return config label indexes

diazandr3s:

### DIFF
--- a/sample-apps/radiology/lib/infers/deepedit.py
+++ b/sample-apps/radiology/lib/infers/deepedit.py
@@ -119,7 +119,9 @@ class DeepEdit(BasicInferTask):
             AsDiscreted(keys="pred", argmax=True),
             SqueezeDimd(keys="pred", dim=0),
             ToNumpyd(keys="pred"),
-            Restored(keys="pred",
-                     ref_image="image",
-                     config_labels=self.labels if data.get("restore_label_idx", False) else None),
+            Restored(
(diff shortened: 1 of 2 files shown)


sample-apps/radiology/lib/infers/deepedit.py
```diff
@@ -119,5 +119,9 @@ def post_transforms(self, data=None) -> Sequence[Callable]:
             AsDiscreted(keys="pred", argmax=True),
             SqueezeDimd(keys="pred", dim=0),
             ToNumpyd(keys="pred"),
-            Restored(keys="pred", ref_image="image"),
+            Restored(
+                keys="pred",
+                ref_image="image",
+                config_labels=self.labels if data.get("client_id", None) == "itk-snap" else None,
```
SachidanandAlle, line 125: whey client-id?  can't we use/define config `restore_label_idx: True/False`

Or what is the behavior for other viewers (ohif/slicer) if this is enabled by default?

SachidanandAlle: At the end, in the response if we are including the label {name vs index} information, any viewer should be able to consume that information and fix the behavior accordingly.

diazandr3s: Agree. Just update that bit

SachidanandAlle (approved):

PR #1533, opened 2023-08-27
Fix network loading

matt3o:
This is the missing code for #1521, where I missunderstood @SachidanandAlle.

The code sets the default value `load_strict=False` of `BasicInferTask` in the derived subclasses. The `BundleInferTask` also now matches the previous behavior.

The only model with a value of True is `HovernetNuclei` . 

This PR fixes the probably breaking network whenever values are missing, since the other PR enabled just that.

### DIFF
--- a/monailabel/tasks/infer/basic_infer.py
+++ b/monailabel/tasks/infer/basic_infer.py
@@ -452,14 +452,17 @@ class BasicInferTask(InferTask):
 
                 if path:
                     checkpoint = torch.load(path, map_location=torch.device(device))
-                    for key in self.model_state_dict:
-                        if key not in checkpoint:
-                            logger.warning(
-                                f"Expected key {key} has not been found in the checkpoint keys: {checkpoint.keys()}"
-                            )
-                            logger.warning("The run will now continue unless load_strict is set to True")
-                            break
                     model_state_dict = checkpoint.get(self.model_state_dict, checkpoint)
+
+                    if set(self.network.state_dict().keys()) != set(checkpoint.keys()):
+                        logger.warning(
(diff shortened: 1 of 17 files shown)


matt3o:
@tangy5 Can you look into this? The way it is some models will probably break in the dev branch since we changed the default value of `load_strict` to `True`

SachidanandAlle:
this is PR now? and why so many changes again? some rebase issue on your branch?

SachidanandAlle (approved):

matt3o:
Hi @SachidanandAlle,

sorry for the confusion. In the other commit I made the error of not fully completing everything, since I wanted to know if the idea is fine for everyone. Therefore I changed the default value in basic_infer and the value for Deepgrow just to show how it would look like. I wanted to add the other inferers afterwards and was waiting for your confirmation, so that I could proceed. However you instantly merged it after the discussion apparently thinking all the work is already done.
Your comments were absolutely right, the way the PR was, it would probably break older models, which is why I opened this second PR to fix that.

This is the final PR, I changed it for all the basic_infer based models I found, I hope that was all of them. About the rebase - yeah this is the code state of the last branch, I did not update the main branch meanwhile. I hope that is fine otherwise I can merge in main, then it should only show the most recent commit

Once again, sorry for the misunderstanding!

SachidanandAlle:
Hey no worries.. I see what was the concern.. by default, we don't load it strict. and hence people were finding difficult to debug why inference results were bad.. (any load failure will have random weights)

so i was ok to make it strict by default.. and override false in each model for people to see it/understand it..

and thanks for the contributions. please continue helping.

Issue #1407, opened 2023-04-17
DeepEdit clicks based refinement seems to be adding segmentation on wrong side


VishalJ99:
Ive taken the default radiology app and used it to train a deepEdit model to do volumetric MRI pancreas segmentation.

The auto segmentation works well however I am experiencing a strange issue when I try and make click based refinements.

Here is an example. This is the auto segmentation generated by the model:
![image](https://user-images.githubusercontent.com/51826812/232453250-e7500fa7-e8f4-4e4e-aae0-168d7cb98590.png)
You can see the right side could use further filling in so i add a foreground click in the desired region and the segmentation is updated to give the following:
![image](https://user-images.githubusercontent.com/51826812/232453678-3af3108d-c1a4-423a-a0b6-2c1348d518fd.png)

The refinement seems to be based on a click that was defined on the left side when I placed it on the right...

Here is the output when I purposefully define the click on the wrong side:

![image](https://user-images.githubusercontent.com/51826812/232454078-0f2ae44b-549d-4e76-8907-21b789ba7bc6.png)

i get the desired output.

I do not experience this issue when making click based refinements on a model trained on the CT spleen dataset downloaded from monai labels example datasets. I have not made any changes to any files other than renaming the labels.

The only explanation i can think of is that my input data is of orientation LAS when the CT spleen data is of orientation RAS. 
However, i noticed that in the training and infer pre transforms, one of the transforms defined is the orientationD transform which re orientates the volumes to RAS, so this should not cause any issue...

VishalJ99:
> DICOM and most (if not all) medical imaging software uses right-handed coordinate system. How did you end up with a left-handed one (LAS)? You may need to resample your image to be in a left-handed coordinate system.

Not sure, however it shouldn't matter because of the orientationD transform defined in the infer and training pre transforms re orientating all volumes to RAS anyway?

lassoan:
It should not matter which right-handed coordinate system you use. If you use a left-handed coordinate system in the image then the image to physical coordinate system transformation matrix will have a negative determinant, which may cause problems in some algorithms. It is also possible that the matrix is orthogonalized somewhere along the way by computing the third axis from the cross product of two other axes, which works perfectly for any right-handed coordinate system but will flip an image axis for images that have left-handed coordinate system.

[...]

PR #1537, opened 2023-08-29
Add transform to apply orientation to guidance

diazandr3s:

### DIFF
--- a/sample-apps/radiology/lib/infers/deepedit.py
+++ b/sample-apps/radiology/lib/infers/deepedit.py
@@ -88,7 +88,7 @@ class DeepEdit(BasicInferTask):
         if self.type == InferType.DEEPEDIT:
             t.extend(
                 [
-                    OrientationGuidanceMultipleLabelDeepEditd(label_names=self.labels),
+                    OrientationGuidanceMultipleLabelDeepEditd(ref_image="image", label_names=self.labels),
                     AddGuidanceFromPointsDeepEditd(ref_image="image", guidance="guidance", label_names=self.labels),
                     Resized(keys="image", spatial_size=self.spatial_size, mode="area"),
                     ResizeGuidanceMultipleLabelDeepEditd(guidance="guidance", ref_image="image"),
--- a/sample-apps/radiology/lib/infers/deepedit.py
+++ b/sample-apps/radiology/lib/infers/deepedit.py
@@ -34,6 +34,7 @@ from monai.transforms import (
 from monailabel.interfaces.tasks.infer_v2 import InferType
 from monailabel.tasks.infer.basic_infer import BasicInferTask
 from monailabel.transform.post import Restored
+from lib.transforms.transforms import OrientationGuidanceMultipleLabelDeepEditd
 
 
 class DeepEdit(BasicInferTask):
@@ -86,6 +87,7 @@ class DeepEdit(BasicInferTask):
         if self.type == InferType.DEEPEDIT:
             t.extend(
                 [
+                    OrientationGuidanceMultipleLabelDeepEditd(key_click='click_coordinates', label_names=self.labels),
                     AddGuidanceFromPointsDeepEditd(ref_image="image", guidance="guidance", label_names=self.labels),
                     Resized(keys="image", spatial_size=self.spatial_size, mode="area"),
                     ResizeGuidanceMultipleLabelDeepEditd(guidance="guidance", ref_image="image"),
--- a/sample-apps/radiology/lib/infers/deepedit.py
+++ b/sample-apps/radiology/lib/infers/deepedit.py
@@ -88,7 +88,7 @@ class DeepEdit(BasicInferTask):
         if self.type == InferType.DEEPEDIT:
             t.extend(
                 [
-                    OrientationGuidanceMultipleLabelDeepEditd(key_click="click_coordinates", label_names=self.labels),
+                    OrientationGuidanceMultipleLabelDeepEditd(label_names=self.labels),
                     AddGuidanceFromPointsDeepEditd(ref_image="image", guidance="guidance", label_names=self.labels),
                     Resized(keys="image", spatial_size=self.spatial_size, mode="area"),
                     ResizeGuidanceMultipleLabelDeepEditd(guidance="guidance", ref_image="image"),
--- a/sample-apps/radiology/lib/infers/deepedit.py
+++ b/sample-apps/radiology/lib/infers/deepedit.py
@@ -11,6 +11,7 @@
 
 from typing import Callable, Sequence, Union
 
+from lib.transforms.transforms import OrientationGuidanceMultipleLabelDeepEditd
 from monai.apps.deepedit.transforms import (
     AddGuidanceFromPointsDeepEditd,
     AddGuidanceSignalDeepEditd,
@@ -34,7 +35,6 @@ from monai.transforms import (
 from monailabel.interfaces.tasks.infer_v2 import InferType
 from monailabel.tasks.infer.basic_infer import BasicInferTask
 from monailabel.transform.post import Restored
-from lib.transforms.transforms import OrientationGuidanceMultipleLabelDeepEditd
 
 
 class DeepEdit(BasicInferTask):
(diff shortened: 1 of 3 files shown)


wyli (approved):
thanks, looks good to me, please note that this only works in 3d.

sample-apps/radiology/lib/transforms/transforms.py
```diff
@@ -505,3 +506,39 @@ def __call__(self, data: Mapping[Hashable, NdarrayOrTensor]) -> Dict[Hashable, N
             if d.get(cache_key) is None:
                 d[cache_key] = copy.deepcopy(d[key])
         return d
+
+
+class OrientationGuidanceMultipleLabelDeepEditd(Transform):
```
yiheng-wang-nv, line 511: Hi @diazandr3s , is there any test case to prove that the oriented coordinates are consistent with oriented volume?

diazandr3s: Hi @yiheng-wang-nv,

I don't have samples for this. But it makes sense to ensure the same orientation is applied to the clicks :)
I also asked @VishalJ99 to check if this works: https://github.com/Project-MONAI/MONAILabel/issues/1407#issuecomment-1697914123

sample-apps/radiology/lib/transforms/transforms.py
```diff
@@ -505,3 +506,39 @@ def __call__(self, data: Mapping[Hashable, NdarrayOrTensor]) -> Dict[Hashable, N
...
+        bs, N = point.shape[:2]
+        point = np.concatenate((point, np.ones((bs, N, 1))), axis=-1)
+        point = rearrange(point, "b n d -> d (b n)")
+        point = affine @ point
+        point = rearrange(point, "d (b n)-> b n d", b=bs)[:, :, :3]
+        return point
+
+    def __call__(self, data):
+        d: Dict = dict(data)
+        if self.key_click == "click_coordinates":
```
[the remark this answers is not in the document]
diazandr3s: Thanks for reviewing, @SachidanandAlle. I agree on this.
I just removed this flag

Nic-Ma:
Hi @diazandr3s ,

@yiheng-wang-nv will update the DeepEdit bundle according to your change here, is it OK? And please help review later.

Thanks in advance.

sample-apps/radiology/lib/transforms/transforms.py
```diff
@@ -505,3 +506,35 @@ def __call__(self, data: Mapping[Hashable, NdarrayOrTensor]) -> Dict[Hashable, N
...
+        bs, N = point.shape[:2]
+        point = np.concatenate((point, np.ones((bs, N, 1))), axis=-1)
+        point = rearrange(point, "b n d -> d (b n)")
+        point = affine @ point
+        point = rearrange(point, "d (b n)-> b n d", b=bs)[:, :, :3]
+        return point
+
+    def __call__(self, data):
+        d: Dict = dict(data)
+        for key_label, val_label in self.label_names.items():
```
yiheng-wang-nv, line 531: `val_label` seems useless here, can use `_` or `for key in label_names.keys()` instead.

BTW, what should be the format of `val_label` in `self.label_names`?  Can you provide an example?

I don't know why label names need to be a dict.

sample-apps/radiology/lib/transforms/transforms.py
```diff
@@ -505,3 +506,35 @@ def __call__(self, data: Mapping[Hashable, NdarrayOrTensor]) -> Dict[Hashable, N
...
+            points = d.get(key_label, [])
+            if len(points) < 1:
+                continue
+            reoriented_points = self.transform_points(
+                np.array(points)[None],
+                np.linalg.inv(d["image_meta_dict"]["affine"].numpy()) @ d["image_meta_dict"]["original_affine"],
```
yiheng-wang-nv, line 537: Hi @diazandr3s , could you switch to use `d["image"].meta["affine"]` and `d["image"].meta["original_affine"]` ?

`"image_meta_dict"` seems outdated and may not update after transforms. The following is my local test code (on MONAI 1.2.0):

```
from monai.transforms import LoadImaged, Orientationd, Compose, EnsureChannelFirstd

image_path = "/workspace/data/medical/Task09_Spleen/imagesTr/spleen_10.nii.gz"

transforms = Compose([
    LoadImaged(keys="image"),
    EnsureChannelFirstd(keys="image"),
    Orientationd(keys="image", axcodes="LPI"),
])

image = transforms({"image": image_path})

print(image["image_meta_dict"]["original_affine"])
print(image["image_meta_dict"]["affine"])
print(image["image"].meta["affine"])
```

Screenshot 2023-09-05 at 7 33 43 PM

BTW, Hi @Nic-Ma @wyli @SachidanandAlle @diazandr3s we may also need to other deepedit related code in MONAI core and MONAILabel if using `"image_meta_dict"`, what do you think?

yiheng-wang-nv: For example, this transform:
https://github.com/Project-MONAI/MONAI/blob/8e77e03105c745936d4ffd98ed536f2bdc132b48/monai/apps/deepedit/transforms.py#L701

SachidanandAlle: @diazandr3s lets get this updated/fixed as per the recommendations.  Let me know if you need any help

diazandr3s: Many thanks for checking this, @yiheng-wang-nv.

I just updated this transform following your suggestions.

The reason for the labels to be a dictionary is that, for some transforms, we need to keep track of the label indexes as well:

`{'spleen': 1, 'right kidney': 2, 'left kidney': 3, 'liver': 6, 'stomach': 7, 'aorta': 8, 'inferior vena cava': 9, 'background': 0}`

This is especially important when the indexes are not contiguous.

yiheng-wang-nv (approved):

sample-apps/radiology/lib/transforms/transforms.py
yiheng-wang-nv: Thanks for the explain @diazandr3s . I think this PR is fine, and for the suggestions of update other transforms, let me create a separate ticket in MONAI core thus I can resolve this comment.

sample-apps/radiology/lib/infers/deepedit.py
```diff
@@ -87,6 +88,7 @@ def pre_transforms(self, data=None):
         if self.type == InferType.DEEPEDIT:
             t.extend(
                 [
+                    OrientationGuidanceMultipleLabelDeepEditd(label_names=self.labels),
```
yiheng-wang-nv, line 91: Hi @diazandr3s , I suggest to add a `ref_image` arg to match up with other transforms.
I did this change in model zoo's PR:
https://github.com/Project-MONAI/model-zoo/pull/501/files#diff-7d1d3c4d9a0b5c7d8e7025ee44af385a0fed3005a63ae4ce68b7f27e639ef7a3R9

diazandr3s: Good idea! Is this what you mean @yiheng-wang-nv?

yiheng-wang-nv: Yes, thanks!

SachidanandAlle:
Do we need this..?

SachidanandAlle (approved):

bastiao:
On our side, we check and it seems that we need this! Thanks for fixing it.

SachidanandAlle:
Closing this due to no activity

bastiao:
@diazandr3s @SachidanandAlle isn't plans to fix/merge this or anything to fix?

joselfrias:
Here is a video demonstrating the issue that @bastiao mentioned, using the DeepEdit model in the Radiology Sample App. When we place positive points to indicate the liver in the top-left area of the image, the model does not appear to apply the feedback correctly. Instead, it produces blue segmentations identifying the liver in the bottom-right area of the image, which is not correct.
We tested with the changes in this PR, and they solved this issue!

https://github.com/user-attachments/assets/400b7c14-cd2c-42d5-a691-2f25323f56b1

SachidanandAlle:
Duplicate PR..

PR #1560, opened 2023-10-09
monai version upgrade to 1.3.0rc3

SachidanandAlle:
1. Upgrade MONAI Core to 1.3.0rc3
2. Fix related to deprecated/delete transforms/attributes
3. Update bundle versions for some example apps/models


wyli (approved):
Thanks, it looks good to me. @diazandr3s please also help confirm the `add_cache_transform`

PR #1561, opened 2023-10-10
Add DICOM SEG support for MONAI Label Inference

joselfrias:
## Context

Currently, in MONAI Label API Inference, and more concretely in a context of a DICOMWeb Datastore, the image inference results are retrieved using the NIFTI format. However, when saving a label, it is possible to trigger MONAI Label to be able to store a DICOM SEG in the image archive using the DICOM-web standard.

[BMDSoftware](https://github.com/BMDSoftware) is integrating MONAI with [Dicoogle](http://www.dicoogle.com/) and found that could be helpful to extend the use of the DICOM standard to MONAI Label Inference, allowing the retrieval of DICOM SEG results, avoiding the need for client applications to convert NIFTI image inference results to the standard DICOM format in the cases where the annotations/labels infered by MONAI Label are not to be saved immediately as final segmentations.

Since MONAI Label already has implemented conversions from NIFTI to DICOM SEG, this PR builts upon this to create DICOM SEG support in MONAI Label Inference.

## Summary

- Adds a new option in MONAI Label Inference API to return a DICOM SEG file.
- Uses existent conversions in MONAI Label to convert from NIFTI to DICOM SEG.

### DIFF
--- a/monailabel/endpoints/infer.py
+++ b/monailabel/endpoints/infer.py
@@ -24,6 +24,7 @@ from fastapi.responses import FileResponse, Response
 from requests_toolbelt import MultipartEncoder
 
 from monailabel.config import RBAC_USER, settings
+from monailabel.datastore.dicom import DICOMWebDatastore
 from monailabel.datastore.utils.convert import binary_to_image, nifti_to_dicom_seg
 from monailabel.endpoints.user.auth import RBAC, User
 from monailabel.interfaces.app import MONAILabelApp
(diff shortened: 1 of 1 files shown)


SachidanandAlle:
> This branch is out-of-date with the base branch

joselfrias:
@SachidanandAlle Thanks for the feedback. We updated the branch. Indeed, it is true that we will need image to copy metadata from headers. If this image does not exist, assuming that the dicom image path is given by the image id in the MONAI datastore, we handle the error. Is there any unhappy path to handle or any other suggestion?

SachidanandAlle:
Only I like to check if data store instance of type Dicomweb.. checking image again is not necessary as Infer has already ran on existing image

SachidanandAlle (approved):

PR #1586, opened 2023-11-04
Add no scribbles model

matt3o:
I don't think this is the best approach to implement this behavior, however, it does work. I am open to implementing this in a completely different way if a maintainer can tell me how else to do this.
As it is this PR adds a new model called "no_scribbles". If you add this model, the scribbles section is deactivated. For an interactive user study I simply wanted to eliminate all possible sources of confusion and errors, since it is click only (interactive DeepEdit style).

### DIFF
--- a/sample-apps/radiology/README.md
+++ b/sample-apps/radiology/README.md
@@ -188,7 +188,8 @@ the model to learn on new organ.
 | Name                 | Values             | Description                                                     |
 |----------------------|--------------------|-----------------------------------------------------------------|
 | use_pretrained_model | **true**, false    | Disable this NOT to load any pretrained weights                 |
-| preload              | true, **false**    | Preload model into GPU                                                                                |
+| preload              | true, **false**    | Preload model into GPU                                          |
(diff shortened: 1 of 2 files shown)


SachidanandAlle:
Instead of adding a model.. 

you can add a flag.. similar to `--conf scribbles=false`

matt3o:
Sweet, thanks for the suggestion! I just reworked it and now it works via `--conf scribbles False`. Default is `True`, so the behavior does not change unless desired.

matt3o:
Is there any documentation somewhere for these kind of flags? Also this is only implemented for the radiology app for now, shall I add it to the other sample-apps as well?

SachidanandAlle:
Looks Good.. for document.. currently I can suggest to add under radiology readme.md and show case this option to enable or disable scribbles

SachidanandAlle (approved):

matt3o:
@SachidanandAlle 
I added the info to the README as discussed above. Also I changed the semantics of the `scribbles` flag to `true` and `false` such that it now matches the other flags 
If all is good from your side, I think the code can be merged

PR #1588, opened 2023-11-07
Update default monailabel URL in OHIF

orenculmination:
Fixing it from http://{host}:8000 to https://{host}, as this is both confusing to users and adds an unnecessary step to the user workflow.


diazandr3s:
> Fixing it from http://{host}:8000 to https://{host}, as this is both confusing to users and adds an unnecessary step to the user workflow.

Hi @orenculmination,

Thanks for the contribution.
Port is needed as MONAI Label by default runs on 8000. Also, port can be modified when starting MONAI Label and OHIF won't connect if port is not correctly specified.

CC @SachidanandAlle

SachidanandAlle:
I am closing this PR for now.  Feel free to reopen with more details

PR #1595, opened 2023-11-20
Mark translatable strings MONAILabel 3D Slicer Extension

DouSam:


plugins/slicer/MONAILabel/MONAILabel.py
```diff
@@ -25,6 +25,8 @@
 import SimpleITK as sitk
 import sitkUtils
 import slicer
+from slicer.i18n import tr as _
```
SachidanandAlle, line 28: is it a good idea to use _?
in python _ is normally used while declaring unused variable names (returns, tuples etc..)

@lassoan ?

lassoan: We were surprised, too, that `_()` is the standard name for translation function in Python, but [that's how it is](https://docs.python.org/3/library/gettext.html), it is a [well-known potential issue](https://peps.python.org/pep-0640/). In practice it does not interfere with returned dummy values because if you overwrite `_` in a function then it only has effect there, so you only need to pay attention on not using `_` for different purposes within a function.

SachidanandAlle:
Can you fix: https://results.pre-commit.ci/run/github/351826770/1700513100.OSh6d9uORu6phJ4FW1r61w

SachidanandAlle:
closed? by accident?

DouSam:
> closed? by accident?

I accidentally closed the PR by fixing the commits without the "Signed-off-by" line, and the two lines with the _ assign.

SachidanandAlle:
pre-commit.ci  fails..
you can reproduce it and fix in your local dev box: 
https://github.com/Project-MONAI/MONAILabel/blob/main/CONTRIBUTING.md#the-contribution-process

SachidanandAlle (approved):

lassoan:
Congrats @DouSam on your first contribution to MONAILabel 🍾

PR #1664, opened 2024-04-01
Add feature "include dicom files"

jose-rfj:
First, I included the tag "model" in the dict "label_info" because I needed the name of the model that generated the label to be further saved.
Then, I included a checkBox in the plugin settings to enable dicom export between server and client. This feature is important because some models can use metadata included in the dicom header, such as slice thickness.
Finally, I added a function call to enable image upload right after segmentation is called, due to the same problem previously described.

### DIFF
--- a/plugins/slicer/MONAILabel/MONAILabel.py
+++ b/plugins/slicer/MONAILabel/MONAILabel.py
@@ -20,6 +20,7 @@ from collections import OrderedDict
 from urllib.parse import quote_plus
 
 import ctk
+import DICOMScalarVolumePlugin
 import qt
 import SampleData
 import SimpleITK as sitk
@@ -1380,6 +1381,14 @@ class MONAILabelWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
             dontShowAgainSettingsKey="MONAILabel/showImageDataSendWarning",
         )
 
+    def get_dicom_files(self):
+        dicom_files = []
+        for path, _, files in os.walk(self.tmpdir):
+            for file in files:
+                if file[-3:] == "dcm":
+                    dicom_files.append(os.path.join(path, file))
+        return dicom_files
+
     def onUploadImage(self, init_sample=True, session=False):
         volumeNode = slicer.mrmlScene.GetFirstNodeByClass("vtkMRMLScalarVolumeNode")
         image_id = volumeNode.GetName()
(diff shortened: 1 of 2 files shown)


SachidanandAlle:
why docs build failing? otherwise changes r ok
@diazandr3s

SachidanandAlle:
can @diazandr3s or @tangy5 run the plugin with these changes.. and see if functionality is working...
also please share some snapshots for this new feature

SachidanandAlle (approved):

jose-rfj:
Dear @diazandr3s 

Thanks for reaching out. I actually updated locally the plugin to store only the first slice, as all of them store the same header, so only one is needed. How do I update that in here? Do I have to create another PR?

Cheers,
Raniery

> Hi @jose-rfj,
> 
> Sorry i'm just catching up on comments/PRs. Many thanks for your contribution.
> 
> I just tested your branch and recorded this video for reference:
> 
>  dicom-files.mp4 
> My question is how could we manage the DICOM files stored in the JSON file? As you see in the video, the MONAI Label updates the JSON file and considers each dcm file as a new volume.
> 
> Additionally, it stores the converted nii.gz file for the DICOM itself.
> 
> Regarding this:
> 
> > This feature is important because some models can use metadata included in the dicom header, such as slice thickness.
> 
> Do you think NRRD firle format will help? What do you think @SachidanandAlle?
> 
> Please let us know,

diazandr3s:
> Dear @diazandr3s
> 
> Thanks for reaching out. I actually updated locally the plugin to store only the first slice, as all of them store the same header, so only one is needed. How do I update that in here? Do I have to create another PR?
> 
> Cheers, Raniery
> 
> > Hi @jose-rfj,
> > Sorry i'm just catching up on comments/PRs. Many thanks for your contribution.
> > I just tested your branch and recorded this video for reference:
> > dicom-files.mp4
> > My question is how could we manage the DICOM files stored in the JSON file? As you see in the video, the MONAI Label updates the JSON file and considers each dcm file as a new volume.
> > Additionally, it stores the converted nii.gz file for the DICOM itself.
> > Regarding this:
> > > This feature is important because some models can use metadata included in the dicom header, such as slice thickness.
> > 
> > 
> > Do you think NRRD firle format will help? What do you think @SachidanandAlle?
> > Please let us know,

Thanks @jose-rfj. 
I guess you can update this PR. No need to create another one.
Let us know,

SachidanandAlle (approved):

PR #1719, opened 2024-07-11
Fix non standard English double-colons

DouSam:


lassoan (approved):
Thank you, this looks good to me.

The motivation for this change was that the `_("Server Error:: Session creation Failed\nPlease upgrade to latest monailabel version (> 0.2.0)"),` message is marked as translatable and translators did not know how to translate the double colon (should just use `:` in other languages or it has some special meaning) - see https://hosted.weblate.org/translate/3d-slicer/monailabel/pt_BR/?checksum=8730a9638b7332b8&q=%3A%3A&sort_by=-priority%2Cposition

It would have been enough to replace `::` by `:` in the translated string, but for consistency it makes sense to avoid using this non-standard typography in all messages that are displayed to users or developers.

lassoan:
@SachidanandAlle could you please check and merge these trivial changes (typos in log messages)? Thank you!

SachidanandAlle (approved):

lassoan:
Thank you!

PR #1740, opened 2024-09-05
Add a test case when opencv is installed, enable both scenario

tangy5:


KumoLiu (approved):
Thanks for the fix, LGTM.

PR #1773, opened 2024-11-04
Minor fixes related to infer and others

SachidanandAlle:
- Fix Docker/GIT Ignore files
- Handle client when Attachment/Image is not sent as part of infer response (JSON Only responses)
- Infer fix get_mimetype and infer_type
- Support `skip_writer` via request param
- Import scribble dependency when it's enabled


diazandr3s (approved):
LGTM

PR #1776, opened 2024-11-06
Use MONAI Label Server directly for cvat annotation plugin

SachidanandAlle:
- Support Latest CVAT
- Use MONAI Label server to run the Inference (all models stay in MONAI Label)
- Fix Interactor/Detector/Tracker for new CVAT
- Update the Readme


diazandr3s (approved):
LGTM

PR #1800, opened 2024-12-16
Update config.py (fix for python 3.9 version)

SachidanandAlle:


lassoan (approved):
It looks good to me.

I've tested it on Python-3.9 by executing `from monailabel.config import settings` and it fixes the error: `AttributeError: 'PathDistribution' object has no attribute 'name'`.

PR #1806, opened 2025-01-15
Enhance task_status to handle job states correctly

bilalcodehub:
Updated the `task_status` function in `cvat.py` to check the state of all jobs within a task. The function now returns "completed" if all jobs are finished or "in_progress" otherwise. This enhancement ensures the function is consistent with CVAT's API behavior.


SachidanandAlle (approved):
Looks good to me

bilalcodehub:
@SachidanandAlle: It seems merging is blocked, likely because I didn’t follow the required commit message format to include author details. What steps can I take at this point to resolve this and move this PR forward?

PR #1811, opened 2025-01-30
Update CellProfiler installation instructions

gnodar01:
Installing cellprofiler from source (which can be a difficult process for some users) is only required if the plugin uses packages not already available within cellprofiler.

`runvista2d.py` uses no external dependencies, so users may instead use the pre-build cellprofiler application, easing installation.


tangy5 (approved):

tangy5:
Hi @gnodar01 , can you push a commit with "-s" in the git commit command to sign-off? it can help pass the DCO, and merge. 
Thanks.

gnodar01:
@ tangy5 done 👍

PR #1854, opened 2025-08-13
Add configuration to allow using itk or pydicom_seg for DICOM SEG

joselfrias:
## Motivation and Context

When working with DICOM images and running inferences to generate DICOM Segs, the ITK tool (`itkimage2segimage`) can sometimes fail: even if MONAILabel successfully produces a segmentation, converting it to a DICOM Seg may abort if a required tag is missing, as shown next, which aborts all the process and in the end no segmentation is returned.

`E: CodingSchemeDesignator (0008,0102) absent in CodeSequenceMacro (type 1)
E: Could not write item #0 in ProcedureCodeSequence: Invalid Value
FATAL ERROR: Writing of the SEG dataset failed! Error: Invalid Value. Please report the problem to the developers, ideally accompanied by a de-identified dataset allowing to reproduce the problem!
ERROR: Conversion failed.
`

To solve this error, a fix is presented here (https://qiicr.gitbook.io/dcmqi-guide/faq#conversion-fails-with-the-missing-attribute-s-error-what-should-i-do), but this is not optimal when dealing with real DICOM data, where files are constantly incoming to the PACS, since it requires to modify the attribute in all dicom files.

Since MONAILabel already supports `pydicom_seg` for DICOM SEG generation, this PR introduces a way to switch between the `itkimage2segimage` tool and `pydicom_seg`, which was previously hardcoded and could not be changed unless we changed the code.

## Summary

- Add `MONAI_LABEL_USE_ITK_FOR_DICOM_SEG` in `monailabel/config.py`, overridable via environment variable. Default is True to preserve existing behavior.
- Update `nifti_to_dicom_seg` in `monailabel/datastore/utils/convert.py` to respect this setting. The function can still override the config if needed; the setting only applies when use_itk is None.

## Test Plan

- Verify that infer requests and save label endpoints work correctly across different datastores.

### DIFF
--- a/monailabel/config.py
+++ b/monailabel/config.py
@@ -115,6 +115,8 @@ class Settings(BaseSettings):
         else "https://huggingface.co/facebook/sam2-hiera-large/resolve/main/sam2_hiera_l.yaml"
     )
 
+    MONAI_LABEL_USE_ITK_FOR_DICOM_SEG: bool = os.environ.get("MONAI_LABEL_USE_ITK_FOR_DICOM_SEG", True)
+
     model_config = SettingsConfigDict(
         env_file=".env",
         case_sensitive=True,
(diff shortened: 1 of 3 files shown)


SachidanandAlle (approved):
Looks good to me

PR #1909, opened 2026-08-10
Assign distinct colors to DICOM SEG segments (ITK export path)

### LINKED ISSUE #1751: Dicom seg shows all labels as color red
**Describe the bug**
When creating a dicom seg object all segments show up as red unless labels are in the 
https://github.com/Project-MONAI/MONAILabel/blob/main/monailabel/datastore/utils/colors.py

**Expected behavior**
Each segment should have different color 

**Proposed fix**
https://github.com/Project-MONAI/MONAILabel/blob/main/monailabel/datastore/utils/convert.py#L98 
change to 
```
out side the loop do 
GENERIC_ANATOMY_COLORS_list = list(GENERIC_ANATOMY_COLORS.values())

        if name in GENERIC_ANATOMY_COLORS:
            rgb = list(info.get("color", GENERIC_ANATOMY_COLORS.get(name, (255, 0, 0))))[0:3]
        else:
            rgb = GENERIC_ANATOMY_COLORS_list[i+1] # picks index from list skipping the background 
```

ousamabenyounes:
## Summary

Fixes #1751 — DICOM SEG shows all segments as the same color (red).

The ITK/dcmqi export path used the same red fallback whenever a segment had no explicit color and its name was absent from `GENERIC_ANATOMY_COLORS`.

## Change

- Add `get_segment_color(name, info, index)` with the existing priority: explicit color, known anatomy color, then an indexed fallback palette color.
- Exclude the black background entry and deduplicate repeated RGB tuples while preserving palette order.
- Use the helper in `_itk_nifti_to_dicom_seg` and cover the first prior duplicate span with 17 unknown segments.

Behavior for explicit colors and known anatomy names is unchanged. The default highdicom path is unaffected.

## Test verification (RED → GREEN)

**RED** — before palette deduplication, the 17-segment regression failed:

```text
AssertionError: 16 != 17
1 failed, 4 passed
```

**GREEN** — after deduplication:

```text
5 passed
monailabel/datastore/utils/colors.py: 100% coverage
```

All repository pre-commit hooks pass locally from a clean isolated cache on Python 3.12. The remote pre-commit.ci run is currently blocked by the pinned `pyupgrade` hook crashing inside pre-commit.ci's Python 3.14 environment (`TypeError: cannot use a bytes pattern on a string-like object`); black, flake8, mypy, and every other remote hook pass.

The full local unit replay is iso-baseline: upstream and this branch both stop during collection on the same pre-existing `girder_client` / `pkg_resources` dependency incompatibility; this branch collects five additional tests before that identical failure.

### DIFF
--- a/monailabel/datastore/utils/colors.py
+++ b/monailabel/datastore/utils/colors.py
@@ -321,3 +321,43 @@ GENERIC_ANATOMY_COLORS = {
     "unknown": (100, 100, 130),
     "cyst": (205, 205, 100),
 }
+
+# Key of the black background entry, excluded from the fallback palette so that a
+# segment never resolves to an invisible (0, 0, 0) color.
+_BACKGROUND_KEY = "background"
+
+# Number of channels in an RGB color triple.
+_RGB_CHANNELS = 3
+
+# Distinct, non-background colors used to give segments without a known anatomy
+# name (and without an explicit color) different colors instead of all defaulting
+# to red. See issue #1751.
+_FALLBACK_COLORS = [color for name, color in GENERIC_ANATOMY_COLORS.items() if name != _BACKGROUND_KEY]
+
+
+def get_segment_color(name, info=None, index=0):
+    """Return an ``[r, g, b]`` display color for a DICOM SEG segment.
+
+    Resolution order:
+
+    1. an explicit ``color`` provided in ``info`` (from the label metadata);
+    2. the palette entry whose key matches ``name`` in ``GENERIC_ANATOMY_COLORS``;
+    3. a palette color chosen by ``index`` so that consecutive segments without a
+       known anatomy name receive *distinct* colors instead of all defaulting to
+       red (see issue #1751).
+
+    Args:
+        name: Segment label used to look up a known anatomy color.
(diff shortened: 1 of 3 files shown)


ousamabenyounes:
Thanks for the review. Addressed the duplicate fallback RGB values in `63bc1e5` by deduplicating the palette while preserving order, and expanded the regression to cover 17 unknown segments.

Validation: the expanded test reproduced RED (`16 != 17`) before the fix, then passed 5/5 with 100% coverage of the changed production file. Pre-commit passed. The full local suite is iso-baseline: upstream and the fixed branch both stop on the same pre-existing `girder_client` / `pkg_resources` collection error.

ousamabenyounes:
Clarification on validation: all repository pre-commit hooks pass locally from a clean isolated cache on Python 3.12. The remote pre-commit.ci status is failing because the pinned `pyupgrade` hook crashes inside pre-commit.ci's Python 3.14 environment with `TypeError: cannot use a bytes pattern on a string-like object`; black, flake8, mypy, and every other remote hook pass. No formatting changes or autofix commit were produced.

